PR #1, opened 2022-06-14
Fix instructions to run npm run start command

adiabramovitch:
Not sure If it is a requirement for mac, but for me running:
`PROXY=true npm run start:beta`

 resulted in: 

```
npm ERR! missing script: start:beta
npm ERR! 
npm ERR! Did you mean this?
npm ERR!     start
```

@amirfefer  Maybe we should add it as a different option?

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ sudo npm run patch:hosts
 
 1. ```npm install```
 
-2. ```PROXY=true npm run start:beta```
+2. ```PROXY=true npm run start``` choose stage and then beta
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +26,7 @@ sudo npm run patch:hosts
 
 1. ```npm install```
 
-2. ```PROXY=true npm run start:beta```
+2. ```npm run start``` choose stage and then beta
```
ezr-ondrej, line 29: I believe you need the PROXY=true for the ui to pull chrome and all the other consoleDot components, don't you?

PR #3, opened 2022-06-20
add github actions

amirfefer:


ezr-ondrej (approved):
Yup, I do not completely understand, but looks reasonable and reasonably similar to other services, lets try it out :+1:

PR #13, opened 2022-06-27
Adding react-query

amirfefer:
This adds [react-query](https://react-query.tanstack.com), it should replace the need to create and manage the global state and simplify the API querying


src/Components/SourcesDropdown/index.js
```diff
@@ -1,58 +1,65 @@
...
+    setChosenSource(id)
+  }
+  const dropdownItemsMapper = () => {
+    if (isLoading)
+      return (
+        <DropdownItem>
+          <Spinner size="md" isSVG />
+        </DropdownItem>
+      );
+    if (sources?.length > 0) {
```
adiabramovitch, line 48: why is there a '?'

adiabramovitch:
looks good! Can you fix the lining? :)

src/Components/SourcesDropdown/index.js
```diff
@@ -1,58 +1,65 @@
...
   const [isOpen, setIsOpen] = React.useState(false);
-  const [sources, setSources] = React.useState([]);
+  const [chosenSource, setChosenSource] = React.useState();
+
+  const {
+    isLoading,
+    error,
+    data: sources,
+  } = useQuery(SOURCES_QUERY_KEY, fetchSourcesList);
+
```
adiabramovitch, line 23: interesting, how doe this work? (a reminder to talk about)

src/Components/SourcesDropdown/index.js
amirfefer: represents [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) :)
if `sources` is null, then `sources.length` will produce a runtime error
the optional chaining stops chaining if theres a null

amirfefer:
@adiabramovitch, linting is fixed :)

PR #24, opened 2022-07-07
Add docs for dev and test

amirfefer:


docs/dev.md
```diff
@@ -0,0 +1,57 @@
+# Development
+This doc explains
```
adiabramovitch, line 2: It seems like the sentence was cut short, is that on purpose?

docs/dev.md
```diff
@@ -0,0 +1,57 @@
+# Development
+This doc explains 
+
+
+## UX / Components
+We use [patternfly](https://www.patternfly.org/v4/) as our primary UX guidance and patternfly-react for our main components store.
+You can find further components and utils which made by `@redhat-cloud-services`
```
adiabramovitch, line 7: the word "which" is not necessary here :)

docs/dev.md
```diff
@@ -0,0 +1,57 @@
...
+      isError,
+      error,
+      data,
+  } = useQuery(MY_QUERY_KEY, fetchSomeList);
+   ```
+   There are plenty of options to choose from and control, by defining a stale query, caching time, manual triggering, pulling and more. 
+   For further reading visit the react-query [docs](https://react-query.tanstack.com/reference/useQuery)
+
+## Client global state
+While the `react-quey` replaces the need for managing the server's data state, we still need from time to time to manage a client global state.
```
adiabramovitch, line 38: missing r: react-query

docs/dev.md
amirfefer: > is that on purpose?

yes, it's still draft :)

ezr-ondrej (commented):
I'm bit missing some documentation on what to store in global state and what in the component state, IMHO we can store most of the data we need locally into our Wizard component :thinking:

amirfefer:
> I'm a bit missing some documentation on what to store in the global state and what in the component state, IMHO we can store most of the data we need locally into our Wizard component 

Propagating many local states deep in the tree could end with a mess  and optimizations issues (each step/substep in the wizard is logically a page that wraps multiple components)
The wizard is expected to grow rapidly (i.e multi-cloud), I was searching for an infrastructure that is simple to consume, well optimized,  and could support the fast-growing.

ezr-ondrej:
> I'd say, as a rule of thumb, if you need a client state in multiple steps, then it makes sense to keep it global (like the native `React.Context` approach)

Thanks, makes sense :+1:

adiabramovitch (dismissed):

amirfefer:
@ezr-ondrej, I updated the global state section, is it clearer now?

ezr-ondrej (approved):
Thanks @amirfefer !

PR #26, opened 2022-07-12
Add review details page

adiabramovitch:
updated
image

### DIFF
--- a/src/Components/Wizard/steps/ReviewDetails/index.js
+++ b/src/Components/Wizard/steps/ReviewDetails/index.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import { Stack, StackItem, Title } from '@patternfly/react-core';
+import { ExpandableAWS } from '../../../ExpandableAWS';
+
+const ReviewDetails = () => {
+  return (
+    <Stack hasGutter>
+      <StackItem>
+        <Title headingLevel="h2">Review Details</Title>
+      </StackItem>
+      <StackItem>
+        <Title headingLevel="h6">
+          Review the information below and click Finish to complete
+          provisioning.
+        </Title>
+      </StackItem>
+      <StackItem>
+        <ExpandableAWS />
+      </StackItem>
+    </Stack>
+  );
+};
+export default ReviewDetails;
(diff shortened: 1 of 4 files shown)


src/Components/Wizard/steps/ReviewDetails/index.js
```diff
@@ -0,0 +1,28 @@
...
+          Review the information below and click Finish to complete
+          provisioning.
+        </Title>
+      </FormGroup>
+      <FormGroup>
+        <ExpandableAWS />
+      </FormGroup>
+      <FormGroup></FormGroup>
+      <FormGroup></FormGroup>
+      <FormGroup></FormGroup>
```
amirfefer, line 24: leftover?

src/Components/Wizard/steps/ReviewDetails/index.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { Form, FormGroup, Title } from '@patternfly/react-core';
+import { ExpandableAWS } from '../../../ExpandableAWS';
+
+const ReviewDetails = () => {
+  return (
+    <Form isWidthLimited maxWidth="80%">
```
amirfefer, line 7: do we need to wrap the review step with a `<Form>`? basically, it's a read-only review with no inputs. no?

amirfefer (commented):
Thanks @adiabramovitch! Looks good, just some minor inline comments

src/Components/Wizard/steps/ReviewDetails/index.js
ezr-ondrej: I tend to agree here :thinking:

adiabramovitch: yes, I agree, so what is it supposed to be? 
Text Input? 
I need to warp the ExpandableAWS and the two titles somehow

ezr-ondrej: Why do you need to wrap? wouldn't empty component `<></>` be enough there?

adiabramovitch: I changed it to Stack and each component inside is inside a Stack item. 
I also updated the screenshot

ezr-ondrej (approved):
Thanks @adiabramovitch ! :+1:

PR #39, opened 2022-07-24
Fix console errors during tests

amirfefer:
While the tests are running there are some console errors:

1. 
```    console.error
      Warning: Each child in a list should have a unique "key" prop.
      Check the render method of `InstanceTypesSelect`. See https://reactjs.org/link/warning-keys for more information.

````
reason: the fixtures don't have `id` props
2.
```
 Warning: You seem to have overlapping act() calls, this is not supported. Be sure to await previous act() calls before making a new one.
```  

reason: `waitFor`  is async and should be wrapped by a promise / async-await


ezr-ondrej (approved):
Thanks @amirfefer  :+1:

PR #40, opened 2022-07-24
Add image prop to wizard

amirfefer:
This adds an `image` prop to the `ProvisioningWizard` component
This prop is an object with a shape of `{ name: string, ami: string }`

The [ami](https://github.com/RedHatInsights/image-builder-frontend/blob/main/src/Components/ImagesTable/ImageLink.js#L29) already exists in the image-builder front application, so it seems redundant to fetch it later in our backend by the image identifier. The image's name prop propagates to the review page, so the user is notified just before the reservation is made.

```js
    <ProvisioningWizard
       isOpen={isWizardOpen}
       onClose={onCloseFn}
       image={{ name: 'example-image', ami: 'ami-0c830793775595d4b' }}
    />
```

- [ ] Testing

### DIFF
--- a/src/Components/Wizard/index.js
+++ b/src/Components/Wizard/index.js
@@ -8,9 +8,9 @@ import APIProvider from '../Common/Query';
 import defaultSteps from './steps';
 import { useDispatch } from 'react-redux';
 
-const ProvisioningWizard = ({ isOpen, onClose, ...props }) => {
+const ProvisioningWizard = ({ isOpen, onClose, image, ...props }) => {
   const [stepIdReached, setStepIdReached] = React.useState(1);
-  const steps = defaultSteps({ stepIdReached });
+  const steps = defaultSteps({ stepIdReached, image });
   const dispatch = useDispatch();
 
   const onNext = ({ id, name }, { prevId, prevName }) => {
@@ -51,8 +51,12 @@ const ProvisioningWizard = ({ isOpen, onClose, ...props }) => {
 };
 
 ProvisioningWizard.propTypes = {
-  isOpen: PropTypes.bool,
+  isOpen: PropTypes.bool.isRequired,
   onClose: PropTypes.func,
+  image: PropTypes.shape({
+    name: PropTypes.string,
+    id: PropTypes.string,
+  }).isRequired,
 };
 
 export default ProvisioningWizard;
--- a/src/Components/DescriptionListAWS/index.js
+++ b/src/Components/DescriptionListAWS/index.js
@@ -1,3 +1,4 @@
+import PropTypes from 'prop-types';
 import React from 'react';
 import {
   DescriptionList,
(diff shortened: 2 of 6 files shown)


ezr-ondrej (commented):
I believe this solution will fall apart once we need any extra information about the image apart of the AMI id.... do you think we can be sure there will be no other information about the image needed?

amirfefer:
we can also propagates an image id for fetching unknown attributes later on 🤔

amirfefer:
This is what the image status endpoint returns from image-builder API: 
```json
{
"image_status": {
  "status": "success",
  "upload_status": {
    "status": "success",
    "type": "aws",
    "options": {
        "ami": "ami-0c830793775595d4b",
        "region": "eu-west-1"
     }
  }, 
"error": {}
}
```
Looks a bit redundant to me to do an async request just for the AMI while it is already available to our consumers.

BTW, there's a nice online `redoc` tool for visualizing image-builder API - https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/osbuild/image-builder/main/internal/v1/api.yaml :)

amirfefer:
After offline discussion, I've created image builder client which can be extended for future multi-cloud usage
https://github.com/RHEnVision/provisioning-backend/pull/116
the consumer should add image object (name and and id):
```js
    <ProvisioningWizard
       isOpen={isWizardOpen}
       onClose={onCloseFn}
       image={{ name: 'example-image', id: 'compose-id' }}
    />
```
The server will verify and pull its `ami` info

src/Components/Wizard/index.js
```diff
@@ -51,8 +51,12 @@ const ProvisioningWizard = ({ isOpen, onClose, ...props }) => {
 };
 
 ProvisioningWizard.propTypes = {
-  isOpen: PropTypes.bool,
+  isOpen: PropTypes.bool.isRequired,
   onClose: PropTypes.func,
+  image: PropTypes.shape({
+    name: PropTypes.string,
+    id: PropTypes.string,
+  }).isRequired,
```
ezr-ondrej, line 59: @Gundersanne does this make sense to you, or shall we get more options from IB? :)
I guess we will rething this once we are in multicloud anyway, so this is M0 options in my mind, but if you have ideas how to make it ready for M1 now, happy to talk it over :)

ezr-ondrej (commented):
needs rebase, but apart of that, it looks ready to me :+1:

ezr-ondrej:
@amirfefer can we have the rebase pls? :(

amirfefer:
rebased, btw we have now a provisioning namespace in stage :)

The wizard testing will be part of https://github.com/RHEnVision/provisioning-frontend/pull/47

ezr-ondrej (approved):
Thanks @amirfefer ! :+1:

PR #42, opened 2022-08-02
Add build and deploy scripts

ezr-ondrej:


deploy/frontend.yml
```diff
@@ -0,0 +1,40 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: provisioning-frontend
+objects:
+  - apiVersion: cloud.redhat.com/v1alpha1
+    kind: Frontend
+    metadata:
+      name: provisioning
```
mshriver, line 9: I think this has to match `provisioning-frontend` resourceTemplate name for `deploy-iqe-cji` to correctly identify the component.

ezr-ondrej: ok, lets try with provisioning-frontend, but this is what I've copied from https://github.com/RedHatInsights/insights-inventory-frontend/blob/master/deploy/frontend.yaml

.rhcicd/pr_check.sh
```diff
@@ -0,0 +1,39 @@
...
+IQE_PLUGINS="hms-provisioning"
+IQE_MARKER_EXPRESSION="smoke"
+IQE_FILTER_EXPRESSION=""
+
+set -exv
+# source is preferred to | bash -s in this case to avoid a subshell
+source <(curl -sSL $COMMON_BUILDER/src/frontend-build.sh)
+BUILD_RESULTS=$?
+
+# Stubbed out for now
```
mshriver, line 30: Yep, @akhil-jha or I will update this to execute tests once we've got some ready. I think this is great to include though so that we're building PR containers.

mshriver (approved):

PR #45, opened 2022-08-05
Name the app

ezr-ondrej:
Set the app name properly in package.json.


ezr-ondrej:
Simple change, going to self-merge to test it, sorry if that causes any inconvenience.
@amirfefer FYI

PR #47, opened 2022-08-07
HMSPROV-155 - trigger aws reservation from UI

amirfefer:


src/API/index.js
```diff
@@ -12,3 +12,7 @@ export const fetchInstanceTypesList = async (sourceId) => {
   );
   return data;
 };
+
+export const createAWSReservation = async (params) => {
+  return axios.post(provisioningUrl('reservations/aws'), params);
```
ezr-ondrej, line 17: are we doing any failure handling here, or is it todo?

adiabramovitch:
I checked with adding hard-coded pubkey id and image id, and it works :) 
should we merge it or do we want to wait for the pubkey page to be implemented? 
What are we going to do regarding the image_id? how do we get it?

src/API/index.js
ezr-ondrej: explained offline, @amirfefer explained we are doing it in sideffect, see below :)

ezr-ondrej:
> What are we going to do regarding the image_id? how do we get it?

In our sample page, we are hardcoding it already I believe, in real world scenario, we should be getting it as a prop to the component from Image builder on component initialization, so that should be fine just by exposing that prop as we do.

> should we merge it or do we want to wait for the pubkey page to be implemented?

I would actually love if we have at least the select first, before merging it :thinking:
But I'd not block on it per se, so if we cant do the select dropdown first, I am fine merging without it :)

src/Components/Wizard/Wizard.js
```diff
@@ -0,0 +1,67 @@
...
+          ? `Provisioning cannot be started due to an error: ${error}`
+          : 'a notification will be trigger after a success or failure',
+      })
+    );
+  };
+
+  // TODO: add pubkey_id
+  const onComplete = () => {
+    mutate({
+      source_id: chosenSource,
```
adiabramovitch, line 37: Can you add the parsing here? 
```js
    var id = parseInt(chosenSource)
    mutate({
      source_id: id,
```

ezr-ondrej: so we actually don't want to parse and keep it a string?

amirfefer:
I'm closing this in favor of https://github.com/RHEnVision/provisioning-frontend/pull/57, the requests will be handle there with a progress screen

PR #52, opened 2022-08-08
Fix npm vulnerabilities

amirfefer:
Running npm audit returns:
```
3 vulnerabilities (2 high, 1 critical)
```

This fixes these  vulnerabilities

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "frontend-starter-app",
+  "name": "provisioning-frontend",
   "version": "0.1.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "frontend-starter-app",
+  "name": "provisioning-frontend",
```
lzap, line 2: Is this intentional? Does not look like a npm only update.

ezr-ondrej: That will be on me I think ... I did update package.json, but did not regenerate the lock, so that change is due from my change :(

ezr-ondrej (approved):
LGTM, thanks :+1:

PR #56, opened 2022-08-17
HMSPROV-195 - UI for uploading new pubkey

amirfefer:
- [x] Add SSH public key step
- [x] Add a new SSH public key form
- [x] Add form validations (simple regex validation for public key format)
- [x] Add wizard step validation (the next button is disabled until the step is validated)
- [x] Add the pubkey name in the review step (with an indicator that it's a new key)
- [ ] unit testing

https://user-images.githubusercontent.com/11807069/185220431-e89670ee-5c73-45a8-970d-cd31a6ad0482.mov

### DIFF
--- a/src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
+++ b/src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
@@ -1,0 +1,119 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Form, FormGroup, TextInput, FileUpload } from '@patternfly/react-core';
+import { useWizardContext } from '../../../Common/WizardContext';
+
+// This is a simple regex format for public ssh key
+const PUBLIC_KEY_FORMAT = '^(ssh|ecdsa)';
+
+const NewSSHKeyForm = ({ setStepValidated }) => {
+  const [wizardContext, setWizardContext] = useWizardContext();
+  const [isLoading, setIsLoading] = React.useState();
+  const [validations, setValidation] = React.useState({
+    sshKey: 'default',
+    sshName: 'default',
+  });
+
+  React.useEffect(() => {
+    // This effect checks if the entire step is validated
+    const errorExists = Object.values(validations).some(
+      (valid) => valid !== 'success'
+    );
+    setStepValidated((prevStepsValidation) => ({
+      ...prevStepsValidation,
+      sshStep: !errorExists,
+    }));
+  }, [validations]);
+
+  const updateValidation = (inputKey, value) =>
+    setValidation((prevValidations) => ({
+      ...prevValidations,
+      [inputKey]: value,
+    }));
+
+  const handleNameChange = (value) => {
+    setWizardContext((prevState) => ({ ...prevState, sshPublicName: value }));
+    if (value.length === 0) {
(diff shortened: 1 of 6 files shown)


adiabramovitch:
According to the mocks, the "Name" and "SSH public key" are supposed to open only after Pressing the "Add and save a new SsH public key"

ezr-ondrej:
> According to the mocks, the "Name" and "SSH public key" are supposed to open only after Pressing the "Add and save a new SsH public key"

Good catch, but I think that can be done in a PR where we are adding the drop-down

src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
```diff
@@ -0,0 +1,113 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Form, FormGroup, TextInput, FileUpload } from '@patternfly/react-core';
+import { useWizardContext } from '../../../Common/WizardContext';
+
+// This is a simple regex format for public ssh key
+const PUBLIC_KEY_FORMAT = 'ssh-[^s]* AAAA[0-9A-Za-z+/,]+ ([^@]+@[^@]+)';
```
ezr-ondrej, line 7: ugh, this does not seem to be correct, the comment is a free text format, that does not require `@` and is optional.
https://gist.github.com/paranoiq/1932126?permalink_comment_id=3211232#gistcomment-3211232 seems to be closer, but I'd be ok keeping the first part and not get too fancy, just prefix `^(ssh|ecdsa)`

@lzap ?

ezr-ondrej (commented):
Looks good, but the regexp needs to be bit different IMHO.

src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
lzap: Yeah, while this is not terribly complex format (I think RFC 4716 covers it) I would definitely disencourage from doing any deep validation in the UI, at least that blocks the user in case it does not match.

If ever, I would just do "frontline" validation in here, for example only ensuring that the field contains three columns separated by whitespace. That will prevent users from accidentally submitting/pasting random text. Something like (pseudo regexp I have zero experience with JS regexp engines):

```
[^s]+\s+[^s]+\s+[^s]+
```

ezr-ondrej: I'd agree, that covering just basic format is enough, tho the last (comment) is optional I believe.

lzap: Oh yeah then we would modify the regexp to something like `[^s]+\s+[^s]+\s*[^s]*` you get the idea.

ezr-ondrej: @amirfefer any progress here? shall we sync up and define this on slack together, or what is your plan here?

amirfefer: updated :)

amirfefer:
@ezr-ondrej, Thanks for the review

I fixed the ssh validation, fixed console warnings regarding form issues 
Tested and works well with #57

PR #64, opened 2022-08-23
Add public SSH key selection

ezr-ondrej:
I've added the `uploadedKey` to the store and changed the validated step.

Error state:
image

### DIFF
--- a/src/Components/Common/WizardContext/initialState.js
+++ b/src/Components/Common/WizardContext/initialState.js
@@ -2,8 +2,11 @@ const initialWizardContext = {
   chosenSource: undefined,
   chosenNumOfInstances: 1,
   chosenInstanceType: undefined,
+  uploadedKey: false,
   sshPublicName: '',
   sshPublicKey: '',
+  chosenSshKeyId: undefined,
+  chosenSshKeyName: null,
 };
 
 export default initialWizardContext;
--- a/src/Components/ExpandableAWS/index.js
+++ b/src/Components/ExpandableAWS/index.js
@@ -3,7 +3,7 @@ import React from 'react';
 import { ExpandableSection } from '@patternfly/react-core';
 import DescriptionListAWS from '../DescriptionListAWS';
 export const ExpandableAWS = ({ imageName }) => {
-  const [isExpanded, setIsExpanded] = React.useState(false);
+  const [isExpanded, setIsExpanded] = React.useState(true);
   const onToggle = (isExpanded) => {
     setIsExpanded(isExpanded);
   };
--- a/src/Components/Wizard/steps/index.js
+++ b/src/Components/Wizard/steps/index.js
@@ -35,7 +35,13 @@ const defaultSteps = ({
   {
     name: 'SSH key authentication',
     id: 4,
-    component: <PublicKeys setStepValidated={setStepValidation} />,
+    component: (
+      <PublicKeys
+        setStepValidated={(validated) =>
+          setStepValidation((prev) => ({ ...prev, sshStep: validated }))
+        }
+      />
+    ),
     canJumpTo: stepIdReached >= 4,
     enableNext: stepValidation.sshStep,
   },
--- a/src/API/index.js
+++ b/src/API/index.js
@@ -6,6 +6,11 @@ export const fetchSourcesList = async () => {
   return data;
 };
 
+export const fetchPubkeysList = async () => {
+  const { data } = await axios.get(provisioningUrl('pubkeys'));
+  return data;
+};
+
 export const fetchInstanceTypesList = async (sourceId) => {
   const { data } = await axios.get(
     provisioningUrl(`sources/${sourceId}/instance_types/`)
--- a/src/API/queryKeys.js
+++ b/src/API/queryKeys.js
@@ -1,4 +1,5 @@
 export const SOURCES_QUERY_KEY = 'sources';
+export const PUBKEYS_QUERY_KEY = 'pubkeys';
 export const instanceTypesQueryKeys = (chosenSource) => [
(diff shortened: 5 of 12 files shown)


src/Components/ExpandableAWS/index.js
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import { ExpandableSection } from '@patternfly/react-core';
 import DescriptionListAWS from '../DescriptionListAWS';
 export const ExpandableAWS = ({ imageName }) => {
-  const [isExpanded, setIsExpanded] = React.useState(false);
+  const [isExpanded, setIsExpanded] = React.useState(true);
```
adiabramovitch, line 6: Does that mean that AWS default is open?

ezr-ondrej: Yes, is that an issue? :thinking: I thought it is better for the time being (until we have other providers)

adiabramovitch:
Do we want to add a test?

adiabramovitch:
And maybe we should change it to SelectOption like instance type? if we want to be consistent

ezr-ondrej:
> Do we want to add a test?

Dunno, IMHO this does nothing and @amirfefer was voicing against too much testing on client, but I'm fine with whatever, I can add test once touching it to change to Select

> SelectOption like instance type

Yes, once that is in, I'll update this one to comply :)

src/Components/Common/WizardContext/initialState.js
```diff
@@ -2,8 +2,11 @@ const initialWizardContext = {
   chosenSource: undefined,
   chosenNumOfInstances: 1,
   chosenInstanceType: undefined,
+  uploadedKey: false,
```
amirfefer, line 5: is this flag needed? if the `chosenSshKeyId` is defined we can deduce it anyway, no?

src/Components/Wizard/steps/index.js
```diff
@@ -35,7 +35,13 @@ const defaultSteps = ({
   {
     name: 'SSH key authentication',
     id: 4,
-    component: <PublicKeys setStepValidated={setStepValidation} />,
+    component: (
+      <PublicKeys
+        setStepValidated={(validated) =>
```
amirfefer, line 40: nice :)

src/Components/Common/WizardContext/initialState.js
```diff
@@ -2,8 +2,11 @@ const initialWizardContext = {
...
   sshPublicName: '',
   sshPublicKey: '',
+  chosenSshKeyId: undefined,
+  chosenSshKeyName: null,
```
amirfefer, line 9: why not use `sshPublicName`?

amirfefer (commented):
Thanks, @ezr-ondrej  👍  looks good
I left some minor comments

src/Components/Common/WizardContext/initialState.js
ezr-ondrej: I don't think so, what happens if you choose a key, but then change your mind and upload it instead, we will have `chosenSshKeyId` defined, won't we?

src/Components/Common/WizardContext/initialState.js
ezr-ondrej: the same reason as above, they could collide, although here we could probably fix that by nullifying the vars on switching, but then you'd loose what you've filled out when you:
1. go to upload a key, you fill out everyting
2. you wanna check you do not have some key that you remembered you might have
3. you switch to a choose public key
4. you go throught them, select some by accident
5. you go back to uploading

IMHO in this scenario the public key should not be gone. I'll write tests for that.

ezr-ondrej: also, I think that `sshPublicName` should be renamed at some point, but that is irrelevant to this discussion :)

src/Components/Common/WizardContext/initialState.js
amirfefer: we can just reset the `chosenSshKeyId` back to undefined when the user uploads a key

src/Components/Common/WizardContext/initialState.js
amirfefer: as you mentioned, this scenario can be fixed by just resetting the state on switching. I'd prefer fewer states, especially when the state is shareable among the whole wizard though I won't block this PR on this, we can improve this later on (i.e better naming `sshPublicName` ->  `uploadedSSHKeyName`)

amirfefer (commented):
Tested and works well, Thanks @ezr-ondrej !
Before merging, what should be the default option? according to the mockups looks like the upload a key is the default option, but I'm not sure

ezr-ondrej:
I believe that selecting is the default, which makes sense, as most users will probably use one (or limited set) key everytime for all the company hosts.
But good question, because it got me thinking about situation you have no keys uploaded, in which case it might be better to default to the uploading.

ezr-ondrej:
Tests added and transformed to using Select component :)

src/Components/Common/WizardContext/initialState.js
ezr-ondrej: I've added test for the scenario I've mentioned above, I believe simple reuse of the same state would break the test now.
Take a look at the test if you agree please :)

src/Components/Common/WizardContext/initialState.js
ezr-ondrej: we can for sure, I just thought that would be much more corner cases to worry about, it is doable, but this felt bit more straight forward :shrug:

amirfefer (approved):
Thanks, @ezr-ondrej ! 
Let's merge!

PR #65, opened 2022-08-24
Add instance type info

adiabramovitch:
Depends on https://github.com/RHEnVision/provisioning-backend/pull/169
imageimage
image

### DIFF
--- a/src/Components/InstanceTypesSelect/index.js
+++ b/src/Components/InstanceTypesSelect/index.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Spinner, FormSelect, FormSelectOption } from '@patternfly/react-core';
+import { Spinner, Select, SelectOption } from '@patternfly/react-core';
 import { useQuery } from 'react-query';
 import { instanceTypesQueryKeys } from '../../API/queryKeys';
 import { fetchInstanceTypesList } from '../../API';
@@ -7,6 +7,7 @@ import { useWizardContext } from '../Common/WizardContext';
 
 const InstanceTypesSelect = () => {
   const [wizardContext, setWizardContext] = useWizardContext();
+  const [isOpen, setIsOpen] = React.useState(false);
   const {
     isLoading,
     error,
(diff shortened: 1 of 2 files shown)


lzap:
I will only comment from the picture on user experience: I would prefer seeing GiBs instead of MiBs since that is how we think about memory. Rounded values with two digits is perhaps the best thing to do, because on AWS there are small instances that have 0.5 and sometimes 0.6 GiB of memory.

Remember to divide by 1024 not 1000 as these are MiBs not MBs :-)

src/Components/InstanceTypesSelect/index.js
```diff
@@ -17,56 +18,70 @@ const InstanceTypesSelect = () => {
...
+      <Select
+        onToggle={onToggle}
+        onSelect={onSelect}
+        isOpen={isOpen}
+        placeholderText="Select instance type"
+        selections={wizardContext.chosenInstanceType}
         aria-label="Select instance type"
-        value={wizardContext.chosenInstanceType}
       >
         {selectItemsMapper()}
```
ezr-ondrej, line 83: I feel like we could do this in-place if we add the ` if (instanceTypes?.length > 0)` for this else branch.
We should probably render something like `Select account to load instances` note on else (if instances are not loaded)

lzap:
Looks like from the UX research that users would like to also see storage info. I think the most important aspect of storage is how much **instance storage** it has. Some types are are EBS-only so we should set this value to 0 and in the UI show something like "EBS-only" too. Since we want the type to be ideally common for all providers, I suggest to use a different term here: SAN (Storage Area Network) which is exactly what EBS is as other vendors have different meanings of the same while SAN is the standard name for this.

I do not want to make this PR complicated than it already is, but if you want, you can implement this it should be as easy as adding a new field (integer perhaps) and setting it to zero if an instance is EBS-only. I think we should definitely avoid adding and kind of SSD/HDD info as vendors have a TON of various configurations and also SAN storage have many categories depending on the connection (fibre channel, iSCSI etc). Let's not spend any time on this for now.

If you feel like it's too much, let me know and let's create an issue for that.

lzap:
_I commented on a different PR, mistake._

src/Components/InstanceTypesSelect/index.js
```diff
@@ -17,56 +18,84 @@ const InstanceTypesSelect = () => {
...
-      </FormSelect>
+      </Select>
+    );
+  } else {
+    return (
+      <React.Fragment>
```
ezr-ondrej, line 89: this could have been simplified
```suggestion
      <>
```

src/Components/InstanceTypesSelect/index.js
```diff
@@ -17,56 +18,84 @@ const InstanceTypesSelect = () => {
...
+        <input
+          className="pf-c-form-control"
+          readOnly
+          type="text"
+          value="Select account to load instances"
+          id="input-readonly"
+          aria-label="Readonly input example"
+        />
+      </React.Fragment>
```
ezr-ondrej, line 98:
```suggestion
      </>
```

ezr-ondrej (approved):
Looks good, let's merge together then :)

Not much testing here yourself :stuck_out_tongue:

adiabramovitch:
> Looks good, let's merge together then :)
> 
> Not much testing here yourself stuck_out_tongue

I have written a test for instance types already (there is nothing to change their because it checks only the length of the array) :D

ezr-ondrej:
:police_officer: ain't happy

ezr-ondrej (approved):
Thanks @adiabramovitch :+1:
I think we can add typeahead now easily, but lets show what we got to UX :)

PR #67, opened 2022-08-28
HMSPROV-208 - Add image selector in demo page

amirfefer:
This adds an image select input in the provisioning demo page:

https://user-images.githubusercontent.com/11807069/187095532-a020470b-719d-49f8-911c-5906914e79f9.mov

### DIFF
--- a/src/Routes/SamplePage/SamplePage.js
+++ b/src/Routes/SamplePage/SamplePage.js
@@ -1,38 +1,96 @@
 import React, { useEffect } from 'react';
 import { withRouter } from 'react-router-dom';
 
-import { Button, StackItem, Stack, Title } from '@patternfly/react-core';
+import {
+  Button,
+  StackItem,
+  Stack,
+  Title,
+  Select,
+  SelectOption,
+  SelectVariant,
+} from '@patternfly/react-core';
 import { Main } from '@redhat-cloud-services/frontend-components/Main';
 import {
   PageHeader,
   PageHeaderTitle,
 } from '@redhat-cloud-services/frontend-components/PageHeader';
+import axios from 'axios';
 
 import './sample-page.scss';
 import ProvisioningWizard from '../../Components/Wizard';
 
 /**
- * A smart component that handles all the api calls and data needed by the dumb components.
- * Smart components are usually classes.
- *
- * https://reactjs.org/docs/components-and-props.html
- * https://medium.com/@thejasonfile/dumb-components-and-smart-components-e7b33a698d43
+This page demonstrates the provisioning UI wizard
  */
+
+// This default AMI value is taken when no image builder api found
+const DEFAULT_AMI = 'ami-05fa00d4c63e32376';
+const IMAGE_BUILDER_URL = '/api/image-builder/v1/composes';
+
 const SamplePage = () => {
   const [isWizardOpen, setWizardModal] = React.useState(false);
+  const [images, setImages] = React.useState([]);
+  const [chosenImage, setChosenImage] = React.useState({
+    id: undefined,
+    name: undefined,
+  });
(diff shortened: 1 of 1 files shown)


lzap:
Amateur UX designer thought: I would swap the UUID and the name. I think humans tend to work better with names than UUIDs ;-) Otherwise it looks good.

src/Routes/SamplePage/SamplePage.js
```diff
@@ -1,38 +1,91 @@
...
+  };
+
+  const onImageSelect = (event, selection, isPlaceholder) => {
+    if (isPlaceholder) return;
+    setChosenImage(selection);
+    setImageSelect(false);
+    console.log('selected image:', selection);
+  };
+
+  const getImageObjet = () => {
```
ezr-ondrej, line 54:
```suggestion
  const getImageObject = () => {
```

src/Routes/SamplePage/SamplePage.js
```diff
@@ -1,38 +1,91 @@
...
+
 const SamplePage = () => {
   const [isWizardOpen, setWizardModal] = React.useState(false);
+  const [images, setImages] = React.useState([]);
+  const [chosenImage, setChosenImage] = React.useState();
+  const [isImageSelectOpen, setImageSelect] = React.useState(false);
 
   useEffect(() => {
+    const fetchImagesFromIB = async () => {
+      const { data } = await axios.get('/api/image-builder/v1/composes');
```
ezr-ondrej, line 35: this is expecting the IB to be deployed alongside provisioning, right?

ezr-ondrej (commented):
We currently do not have Image builder in ephemeral env, so merging this would break ephemeral environment, right?
Could we implement a workaround for that?
My idea would be to allow a failover to filling out the ami of the image directly (which is now supported by API: https://github.com/RHEnVision/provisioning-backend/pull/184)

amirfefer:
@lzap 
> Amateur UX designer thought: I would swap the UUID and the name. I think humans tend to work better with names than UUIDs ;-) Otherwise it looks good.

I agree :) though this is just for us, a demo page to demonstrate the wizard itself

amirfefer:
@ezr-ondrej - I have added a default AMI when the IB api fails
@lzap - I switched between the name and id values for better UX

src/Routes/SamplePage/SamplePage.js
```diff
@@ -1,38 +1,96 @@
...
- * A smart component that handles all the api calls and data needed by the dumb components.
- * Smart components are usually classes.
- *
- * https://reactjs.org/docs/components-and-props.html
- * https://medium.com/@thejasonfile/dumb-components-and-smart-components-e7b33a698d43
+This page demonstrates the provisioning UI wizard
  */
+
+// This default AMI value is taken when no image builder api found
+const DEFAULT_AMI = 'ami-05fa00d4c63e32376';
```
ezr-ondrej, line 28: problem with preset ami is, that it has to be shared with the account you're about to deploy into, so this image is deployable only to your account if I'm not mistaken. so allowing to fill out the ami manually would be quite awesome for QE, without it everyone would have to use your account to test with in ephemeral.

amirfefer: This `ami` is just example. it's not mine :)

Do we actually want QE test this demo page? it's not part of the production
but we can add an hidden input so QE can edit it during tests

ezr-ondrej: not just QE, but anyone in the ephemeral environment.
I mean without adding any input, QE (nor devs) can trigger the real deployment, because there is no other way to provide an image.

ezr-ondrej (approved):
OTH this is quite nice improvement on itself, I can play with taking it further myself :)

PR #70, opened 2022-08-29
Add storage info

adiabramovitch:
image

### DIFF
--- a/src/API/index.js
+++ b/src/API/index.js
@@ -8,7 +8,7 @@ export const fetchSourcesList = async () => {
 
 export const fetchInstanceTypesList = async (sourceId) => {
   const { data } = await axios.get(
-    provisioningUrl(`sources/${sourceId}/instance_types/`)
+    provisioningUrl(`sources/${sourceId}/instance_types`)
   );
   return data;
 };
--- a/src/Components/InstanceTypesSelect/index.js
+++ b/src/Components/InstanceTypesSelect/index.js
@@ -43,15 +43,17 @@ const InstanceTypesSelect = () => {
       <SelectOption
         aria-label={'Instance Type item'}
         key={instanceType.id}
-        description={
-          instanceType.cores +
-          ' cores | ' +
-          instanceType.vcpus +
-          ' vCPU | ' +
-          (parseFloat(instanceType.memory) / 1024).toFixed(1) +
-          ' GiB memory | ' +
-          instanceType.architecture
-        }
+        description={`${instanceType.cores} cores | 
+          ${instanceType.vcpus} vCPU | 
+          ${(parseFloat(instanceType.memory_mib) / 1024).toFixed(
+            1
+          )} GiB memory | 
+          ${
+            instanceType.storage_gb > 0
+              ? instanceType.storage_gb + ' GB storage | '
+              : 'EBS - only | '
+          }
+          ${instanceType.architecture}`}
         value={instanceType.name}
       />
     ));


src/API/index.js
```diff
@@ -8,7 +8,7 @@ export const fetchSourcesList = async () => {
 
 export const fetchInstanceTypesList = async (sourceId) => {
   const { data } = await axios.get(
-    provisioningUrl(`sources/${sourceId}/instance_types/`)
+    provisioningUrl(`sources/${sourceId}/instance_types`)
```
adiabramovitch, line 11: got 404 with the / so I removed it

src/Components/InstanceTypesSelect/index.js
```diff
@@ -43,15 +43,18 @@ const InstanceTypesSelect = () => {
...
         aria-label={'Instance Type item'}
         key={instanceType.id}
-        description={
-          instanceType.cores +
-          ' cores | ' +
-          instanceType.vcpus +
-          ' vCPU | ' +
-          (parseFloat(instanceType.memory) / 1024).toFixed(1) +
-          ' GiB memory | ' +
+        description={''.concat(
```
amirfefer, line 46: why not use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) instead of concat?

amirfefer (commented):
Looks good to me :)

ezr-ondrej (dismissed):
Might be cleaner with templating, I don't know what platform insights targets, but I guess we are safe with those :)
I myself will not block it on that tho :)

src/Components/InstanceTypesSelect/index.js
```diff
@@ -43,15 +43,17 @@ const InstanceTypesSelect = () => {
...
+          ${instanceType.vcpus} vCPU | 
+          ${(parseFloat(instanceType.memory_mib) / 1024).toFixed(
+            1
+          )} GiB memory | 
+          ${
+            instanceType.storage_gb > 0
+              ? instanceType.storage_gb + ' GB storage | '
+              : 'EBS - only | '
+          }
+          ${instanceType.architecture}`}
```
ezr-ondrej, line 56: Hehe, in this case templating actually looks worse IMHO, but I don't care much I'm good either way, but seeing it it just looks ugly xD

ezr-ondrej (approved):
QE tooling seems to be broken, failures are not related to this PR

src/Components/InstanceTypesSelect/index.js
adiabramovitch: I agree, Should I change it back?

amirfefer: yeah... I'd clean it by extracting variables outside to improve readability, something like
```js
const typeStorage =  instanceType.storage_gb > 0
              ? `${instanceType.storage_gb}  GB storage |`
              : 'EBS - only | '

const typeMemory= `${parseFloat(instanceType.memory_mib) / 1024).toFixed(1))} `


description=`${instanceType.cores} cores | 
                       ${instanceType.vcpus} vCPU |
                       ${typeStorage}
                        ${typeMemory} GiB memory |`

```

ezr-ondrej: agreed, but I guess we can do that in a follow-up :)
Lets merge so anyone can continue improving :)

ezr-ondrej (approved):
Thanks @adiabramovitch ! :+1:

PR #71, opened 2022-08-30
Add instance types filter

amirfefer:
I am tired of manually searching `t1.micro` all the time 🤣 

https://user-images.githubusercontent.com/11807069/187408684-e45baf67-ed41-4dfe-9d6d-ec6971451cb1.mov

### DIFF
--- a/src/Components/InstanceTypesSelect/InstanceTypesSelect.test.js
+++ b/src/Components/InstanceTypesSelect/InstanceTypesSelect.test.js
@@ -13,6 +13,8 @@ describe('InstanceTypesSelect', () => {
 
 const mountSelectAndClick = async () => {
   render(<InstanceTypesSelect />);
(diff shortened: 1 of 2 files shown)


ezr-ondrej (commented):
Awesome, thanks for doing that! :)
Although I wonder, in PF guide [`typeahead`](https://www.patternfly.org/v4/components/select#typeahead) is done by only the `variant=SelectVariant.typeahead`, do we need a custom filter function? PF seems to have implemented a sane default, can't we just use that?

ezr-ondrej (dismissed):
Works, thanks @amirfefer ! :+1:

ezr-ondrej:
Oh the test is now failing :(

amirfefer:
so let's use the filter input for now, tests are green

ezr-ondrej (approved):
Awesome addition, thanks @amirfefer ! :tada:

PR #72, opened 2022-08-30
Allow entering AMI manually from sample page

ezr-ondrej:
Just a simple addition of input for manually entering an AMI that we want to provision.
This should be just a fallback, when IB is not available, but will be useful until IB is in ehpemeral environment and also for easier local testing.

### DIFF
--- a/src/Routes/SamplePage/SamplePage.js
+++ b/src/Routes/SamplePage/SamplePage.js
@@ -6,6 +6,8 @@ import {
   StackItem,
   Stack,
   Title,
+  FormGroup,
+  TextInput,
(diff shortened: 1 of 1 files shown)


adiabramovitch:
LGTM except for the size of the select bar but it is not important 😄

ezr-ondrej:
> LGTM except for the size of the select bar but it is not important smile

Yeah, it is bit oversized, but I was lazy to look for a solution xD
And given its just for testing...

amirfefer (approved):
LGTM, thanks @ezr-ondrej !

PR #74, opened 2022-08-30
remove unused dependencies

amirfefer:
This PR removes `@patternfly/react-table`
I wonder if we want also to remove redux if we are not going to use toast notifications


ezr-ondrej:
> I wonder if we want also to remove redux if we are not going to use toast notifications

I'm all for removing everything we are not using :+1:

ezr-ondrej (approved):
Thanks @amirfefer ! :+1:

PR #80, opened 2022-08-31
Fix new public key validations

amirfefer:


ezr-ondrej (approved):
Sorry for causing this and thanks for fixing this @amirfefer ! :tada:

I wonder if everyone who does refactoring of something that is not tested, shall write tests for the refactored component (in this case I should have written tests for this while touching it).

PR #89, opened 2022-09-02
revert react-router-dom bump

ezr-ondrej:


lzap:
This is the reason why I opened the thread this morning on Slack (before we actually realized the UI was broken). We could only monitor security notifications generated by GitHub and upgrade only the most important and relevant dependencies.

There is no rush in onboarding latest and greatest deps on a daily basis, in fact, it's for the worse as we can see here. In backend, we tend to update all dependencies once per month on average.

PR #90, opened 2022-09-04
HMSPROV-166 - clear the wizard on close

amirfefer:
This addition unmounts the wizard component and its context after the wizard is closed.


ezr-ondrej (commented):
Won't this keep the state of `stepIdReached` and `stepValidation`? :thinking:
I guess that we can just not render the component at all in the sample page and that would clean up the entire state?

amirfefer:
Yep, you're right
We can unmount the wizard from the sample page, but it is just a usage example, a control wizard is better, protecting the wizard from consumers' mistakes :)

PR #91, opened 2022-09-05
Disable dependabot

ezr-ondrej:
Keep only security alerts.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,4 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


amirfefer (commented):
Do we want to add an allow list,  so we won't be behind patternfly and console dot shared packages?

ezr-ondrej:
> Do we want to add an allow list, so we won't be behind patternfly and console dot shared packages?

Sure, done :)
I still believe limiting the number of PRs will make it more actionable, so keeping just 2 :)

amirfefer (approved):
Thanks @ezr-ondrej !

PR #92, opened 2022-09-05
Add region to createAWSDeployment

adiabramovitch:
This is a temporary fix!
- Removing the hard-coded "US- east" 
- Using the default value we have in the wizardContext. 
After getting all the information we need about regions from image builder we would list only the regions the customer had chosen in image builder step.


ezr-ondrej (approved):
`us-east-1` is probably not what we want to show to the user, but I guess we can live with it now :)
Thanks @adiabramovitch ! :+1:

adiabramovitch:
> `us-east-1` is probably not what we want to show to the user, but I guess we can live with it now :) Thanks @adiabramovitch ! +1

I thought about doing a mapper but I thought it is too soon for now :)

PR #95, opened 2022-09-12
Allow editing of the Uploaded pubkey

ezr-ondrej:
This allows editing of the key or inserting it manually.

### DIFF
--- a/src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
+++ b/src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
@@ -47,13 +47,29 @@ const NewSSHKeyForm = ({ setStepValidated }) => {
       updateValidation('sshKeyBody', 'error');
     }
   };
+  const handleSSHKeyText = (text) => {
+    setWizardContext((prevState) => ({
+      ...prevState,
+      sshPublicKey: text,
+    }));
+    if (text === '') {
+      updateValidation('sshKeyBody', 'default');
+    } else {
+      updateValidation(
+        'sshKeyBody',
+        validatePublicKey(text) ? 'success' : 'error'
+      );
+    }
+  };
 
   const handleClear = () => {
     setWizardContext((prevState) => ({
       ...prevState,
-      sshPublicKey: undefined,
       sshPublicName: undefined,
+      sshPublicKey: undefined,
     }));
+    updateValidation('sshKeyName', 'default');
+    updateValidation('sshKeyBody', 'default');
   };
 
   const handleFileReadStarted = () => {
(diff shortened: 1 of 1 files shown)


src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
```diff
@@ -95,12 +118,14 @@ const NewSSHKeyForm = ({ setStepValidated }) => {
         <FileUpload
           id="public-key-value"
           onDataChange={handleSSHKeyChange}
-          allowEditingUploadedText={false}
+          allowEditingUploadedText={true}
```
amirfefer, line 121: nit: you can remove the `true` assignment :)

src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
```diff
@@ -95,12 +118,14 @@ const NewSSHKeyForm = ({ setStepValidated }) => {
...
+          onTextChange={handleSSHKeyText}
```
amirfefer, line 122: can't we use the `handleSSHKeyChange`? looks similar to me?

amirfefer (commented):
Thanks @ezr-ondrej, looks good to me, just a few nits

src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
ezr-ondrej: false is default I believe, at least according to docs :)

src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
ezr-ondrej: they differ, we do not want to upload invalid key, but we want to allow adding more text to a field that is not valid yet :)

src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
amirfefer: I mean you don't need to specify the truth value in react component:
```js
<Component thruthyProp /> 
// equals 
<Component thruthyProp={true} />
```

ezr-ondrej: Ah! now I got it, sure it will look much better :)

src/Components/Wizard/steps/Pubkeys/NewKeyForm.js
ezr-ondrej: it might be similar and possible to join these methods, but I personally would introduce more bugs then if I write two xD

amirfefer (approved):
Thanks @ezr-ondrej  👍

PR #100, opened 2022-09-13
HMSPROV-242 - add reservation polling

amirfefer:
https://user-images.githubusercontent.com/11807069/190016430-3f1423bb-b7f6-4c6c-b0b7-9af8fcd67d0f.mov

PR #101, opened 2022-09-14
HMSPROV-282 - add back button in progress step when error occurred

amirfefer:
on top of #100 

This adds a back button if there is any error during deployment

https://user-images.githubusercontent.com/11807069/190150728-48484baa-6afd-4ded-bdd6-864857d895b0.mov


ezr-ondrej (approved):
Awesome stuff ^_^

PR #106, opened 2022-09-27
Remove Azure and Google steps

ezr-ondrej:
Azure and Google are not used and we need dynamic strategy for adding 
these in the future anyway, so just removing for now seems the best 
strategy.

Also renames the Wizard component directory to match the component name.


lzap (approved):
Will do my best reviewing these.

lzap:
Merged.

PR #108, opened 2022-10-02
HMSPROV-290 - fix validations after clicking back

amirfefer:
https://user-images.githubusercontent.com/11807069/193471364-1483db08-b01d-428c-a700-8841b52c0f0d.mov


ezr-ondrej (approved):
I'd like to be able to also pass the validations, but this fixes it for this step, lets get it in and imrpove on when there's time.

PR #115, opened 2022-10-05
Improve instance type select

ezr-ondrej:
Limit height to allow scroll on select items.
Add limit for items loaded at once to incentives usage of search.

Refs HMSPROV-292

[Screencast from 7.10.2022 13:01:30.webm](https://user-images.githubusercontent.com/2884324/194540628-f3e0d126-55b4-4d00-9cf1-1696a849e58e.webm)

### DIFF
--- a/src/Components/InstanceTypesSelect/index.js
+++ b/src/Components/InstanceTypesSelect/index.js
@@ -6,9 +6,14 @@ import { instanceTypesQueryKeys } from '../../API/queryKeys';
 import { fetchInstanceTypesList } from '../../API';
 import { useWizardContext } from '../Common/WizardContext';
 
+const OPTIONS_PER_SCREEN = 5;
+
 const InstanceTypesSelect = ({ setValidation }) => {
   const [wizardContext, setWizardContext] = useWizardContext();
   const [isOpen, setIsOpen] = React.useState(false);
+  const [numOptions, setNumOptions] = React.useState(OPTIONS_PER_SCREEN);
+  const [filteredTypes, setFilteredTypes] = React.useState(null);
+  const [prevSearch, setPrevSearch] = React.useState('');
   const {
     isLoading,
     error,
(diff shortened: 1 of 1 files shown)


lzap:
Upward dropdown looks weird to me, it's like driving on the left side of the road :-) I am not against it.

What I feel like a lacking feature is to be able to search through the dropdown, we will present around 300 items there typically that can be a pain to find an entry via mouse.

Nit: "EBS only" and not "EBS - only" ?

ezr-ondrej:
I'll need @MariSvirik to chip in on the up (I agree with that being weird, but it does not scroll! Win? xD)

I can say that after trying driving on the left side for a week it always felt bit more natural for some reason.
Just sayin you have to try for a while and MAYBE it can be better once you get used to it xD 

> What I feel like a lacking feature is to be able to search through the dropdown, we will present around 300 items there typically that can be a pain to find an entry via mouse.

We do have typeahead there (you can type in the select and it filters the options as you type).

MariSvirik:
@lzap  We use typeahead - that's what AWS has and seems to be enough
I would go with dropdown going down (not up)

About scroll: 
1. One option to kill forever scroll: [View more](https://www.patternfly.org/v4/components/select/html/#view-more-menu-item) with a limited length of the menu. 
Screenshot 2022-10-06 at 12 35 29

2. The second option is just to limit the length of the menu. The menu can exceed the length of the modal.

My suggestion in the [doc](https://docs.google.com/document/d/1czymMwNY4Tn0BtyLeYQTwroowwqcvODW3FHStkwD5_I/edit#) was to scroll and make the dropdown sticky at the top of the page so you can see the majority of the dropdown and menu items - and then scroll just in the menu. But not sure if that's easily feasible. 

Screenshot 2022-10-06 at 12 46 54

ezr-ondrej:
I've tried to go with the down and max height, so you scroll the wizard only a bit and then you continue in the select, not great, but kinda works.

I like the view more, take a look at new screens pls @MariSvirik if you're against, we can try to invest into load more :) but I think major use case here is filtering the instances by typing... (maybe view more is best at that usecase anyway?)

ezr-ondrej:
Only outstanding question now from UX is whether the select options should not cover the wizard buttons.
This needs more investigation and has been offloaded to https://issues.redhat.com/browse/HMSPROV-312

PR #116, opened 2022-10-05
Review step ux

ezr-ondrej:
Unify finish button text and use "Launch" label

Add exit confirmation, when user has reached the review step and thus 
filled a lot of data (previous steps will still allow canceling without confirm).

exit_dialog

### DIFF
--- a/src/Components/ConfirmModal/index.js
+++ b/src/Components/ConfirmModal/index.js
@@ -1,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button, Modal, ModalVariant } from '@patternfly/react-core';
+
+const ConfirmModal = ({ isOpen, onConfirm, onCancel }) => (
+  <Modal
+    ouiaId="app-confirm-modal"
+    id="app-confirm-modal"
+    aria-label="confirm cancel of launch modal"
+    variant={ModalVariant.small}
+    title="Exit instance launching?"
+    isOpen={isOpen}
+    onClose={onCancel}
+    actions={[
+      <Button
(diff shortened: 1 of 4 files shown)


lzap:
I'd probably say something like "Leave image launch wizard?" but this is on Maria.

ezr-ondrej:
Tagging @MariSvirik

MariSvirik:
Contacted doc team. I would run it past a native speaker.
I would personally go with the microcopy that Ondrej wrote: 
1. PF4 uses "exit" as a verb.  
2. @ezr-ondrej add a period at the end of the sentence.
3. @lzap using "wizard" in microcopy is not a good practice. Who or what is a wizard?
4. I would even use the blue primary button. I assume it's not a super destructive action. But that's not very important.

Example from sources
Screenshot 2022-10-06 at 12 09 00
If we already decided on using Launch everywhere (and it's approved by PMs), let's use "Launch"

MariSvirik:
And add a toast after the user exit the flow successfully.
Example from c.rh.c
Screenshot 2022-10-06 at 12 30 14

bangelic:
I agree with @MariSvirik and her comments but I don't know how often I have ever seen "Stay" as an option even though I am not against using it. "Cancel" is usually the option I see when the user decides they do not want to exit. But I see no problem using "Stay". As she said, use a period at the end of "All inputs will be discarded." otherwise this looks good and makes sense to me.

ezr-ondrej:
Given I've implemented all required, I'm merging, this will go through final ack of UX, so if this is still not good enough, we will improve later on :)

PR #125, opened 2022-10-11
fix(pubkeys): Component refresh does not move cursor

ezr-ondrej:
Parent component refresh makes cursor to move.
We need to keep the state in the local component

Fixes HMSPROV-270


lzap (approved):
No idea what you are talking about, but it works for me.

PR #126, opened 2022-10-11
UX(finish): Progress step UX

ezr-ondrej:
Use correct icons, improve copywrites.

Screenshot from 2022-10-11 18-03-17
[Screencast from 11.10.2022 18:06:24.webm](https://user-images.githubusercontent.com/2884324/195152596-ba5bd573-2fd0-4d54-a9a1-97be241a3d23.webm)

### DIFF
--- a/src/Components/ProvisioningWizard/steps/FinishProgress/index.js
+++ b/src/Components/ProvisioningWizard/steps/FinishProgress/index.js
@@ -10,7 +10,11 @@ import {
   Button,
   WizardContextConsumer,
 } from '@patternfly/react-core';
-import { CogsIcon, OkIcon, ErrorCircleOIcon } from '@patternfly/react-icons';
+import {
+  CogsIcon,
+  CheckCircleIcon,
+  ExclamationCircleIcon,
+} from '@patternfly/react-icons';
 import { useWizardContext } from '../../../Common/WizardContext';
 import { useMutation, useQuery } from 'react-query';
 import {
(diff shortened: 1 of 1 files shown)


lzap:
Just an observation: errors coming out from the backend will always be lowercase (non-sentences). So you could format this as: An error occured: reason from the backend.

But this can be done later as @avitova is working on error payload improvements.

src/Components/ProvisioningWizard/steps/FinishProgress/index.js
```diff
@@ -117,31 +121,26 @@ const FinishStep = ({ onClose, imageID }) => {
     : false;
```
avitova, line 121: Do we have to use the ternary operator? Would possibly `polledReservation && polledReservation.success === false` not work? I've been only working in typescript if that makes a difference.

ezr-ondrej: maybe even using the conditional operator `polledReservation?.success === false` would work here :)
It's already existing code, do you want to send a patch, or shall we include it here? :)

avitova: Let's include this here since you already modify the file in this PR. :)

src/Components/ProvisioningWizard/steps/FinishProgress/index.js
```diff
@@ -167,15 +165,15 @@ const FinishStep = ({ onClose, imageID }) => {
...
           <EmptyStateSecondaryActions>
-            {isError && (
-              <Button onClick={() => goToStepById(1)} variant="link">
-                Back
-              </Button>
-            )}
             <Button
               variant="link"
-              isDisabled={!showCloseButton()}
+              isDisabled={!isError && activeStep < 2}
```
amirfefer, line 176: what if there's an error during step 2 and above? shouldn't we have a clickable close button?

amirfefer (approved):
Tested and seem to work well
thanks @ezr-ondrej , @avitova !

PR #134, opened 2022-10-17
Add documentation on how to run apps together

ezr-ondrej:


lzap:
Do you want me to test this? Is this ready?

ezr-ondrej:
Yeah, this is ready :) but testing it requires multiple repositories, so I guess @adiabramovitch or @amirfefer can set this up, in case they might want to work on IB's frontend improvements :)

README.md
```diff
@@ -37,8 +37,48 @@ sudo npm run patch:hosts
...
+
+The guide for running it is at https://github.com/RedHatInsights/cloud-services-config#testing-your-changes-locally.
+
+To run with another app you can easily run one app just for serving static assets and the other as the main app.
+Pointing the asset route to the statically served assets of the second app.
+
+#### Image builder with Provisioning static assets
+
+
+Run Provisioning frontend with `npm run start:federated`.
```
amirfefer, line 53: it fails on my env without specifying `FEC_CONFIG_PATH`, should we add `--config` flag with the config path?

README.md
```diff
@@ -37,8 +37,48 @@ sudo npm run patch:hosts
...
+     : '/insights/image-builder',
++  routes: {
++    '/apps/provisioning': { host: "http://localhost:8003" },
++    '/api/provisioning': { host: "http://localhost:8000" },
++    '/config/chrome': { host: 'http://localhost:8889' }
++  }
+   };
+```
+
+Run image builder by `STAGE=true npm prod-stable`.
```
amirfefer, line 76: `run` is missing:
```
STAGE=true npm run prod-stable
```

amirfefer (commented):
It works for me, just a few nitpicks

BTW, the local installation of `cloud-services-config` tutorial uses `beta`, don't we want to be consistent and use `prod-beta` in IB  as well during local dev?

README.md
ezr-ondrej: it should have been fixed by https://github.com/RedHatInsights/frontend-components/pull/1631 I guess we should bump our dependency and then it would be fine? but I'm fine adding the config flag :)

ezr-ondrej:
The cloud-config runs in all envs under the same instance, it only depends wheter the request for config goes to /beta/chrome/config or just /chrome/config, but the same instance is serving it :thinking:

amirfefer (approved):
LGTM :)

PR #137, opened 2022-10-18
Add ouia for pubkey

mshriver:
Ready for review, I've checked the deployed image.

FileUpload doesn't actually support it, happy to remove that one if reviewers would rather it not be there until its supported.


amirfefer (approved):
Thanks @mshriver 👍

PR #138, opened 2022-10-19
Align slider max label

amirfefer:
Screen Shot 2022-10-19 at 20 02 29

### DIFF
--- a/src/Components/InstanceCounter/index.js
+++ b/src/Components/InstanceCounter/index.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import { Slider } from '@patternfly/react-core';
 import { useWizardContext } from '../Common/WizardContext';
+import './styles.css';
+
 const MAX_INSTANCES = 45;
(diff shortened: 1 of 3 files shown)


ezr-ondrej (approved):
How did you convinced the styles to load up?

src/Components/InstanceCounter/index.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import { Slider } from '@patternfly/react-core';
 import { useWizardContext } from '../Common/WizardContext';
+import './styles.css';
```
amirfefer, line 4: This import works well, it just wraps the css classes with a `provisioning` class
For example, if we define a new class `.example` it built as
```css
.provisioning .example {
  width: auto
}

therefore I added a `provisioning` class to our wizard down below

amirfefer:
@ezr-ondrej I've  just explained above :)

PR #141, opened 2022-11-02
Added ouiaID and id for launch page

akhil-jha:


src/Components/ProvisioningWizard/steps/FinishProgress/index.js
```diff
@@ -96,7 +96,7 @@ const FinishStep = ({ onClose, imageID }) => {
     title = 'Launching of system(s) had failed';
     iconProps = { color: pf_danger_color_100, icon: ExclamationCircleIcon };
   } else if (activeProgress === 100) {
-    title = 'Launching of system(s) had successeded';
+    title = 'Launching of system(s) had succeeded';
```
mshriver, line 99: `has` is probably more grammatically relevant here for all of these titles.

`Launching of system(s) has failed`
`Launching of system(s) has succeeded`

And we can probably drop a few words entirely and just state `Launching system(s): Success`

mshriver (approved):

src/Components/ProvisioningWizard/steps/FinishProgress/index.js
ezr-ondrej: I'd rather leave this up to UX/docs, so far the guidance usually is that using full sentences is much better and these shortcuts are usually not worth it.
But we can change it once we get docs opinion on the matter :)

ezr-ondrej (approved):

PR #151, opened 2022-11-15
enable running front dev with staged api backend

amirfefer:
This PR extracts the local routes (i.e local API server) from the `fec.config` and allows running frontend with no local provisioning API (communicates directly to the stage)

``` bash
# runs the provisioning-frontend without local provisioning API
npm run start:stage 
```

```bash
# runs the provisioning-frontend including webpack proxy for communicating with the local provisioning API
npm run start
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ yarn-error.log*
 
 .DS_Store
 coverage
+.cache/*
--- a/fec.config.js
+++ b/fec.config.js
@@ -14,9 +14,7 @@ module.exports = {
    * Add additional webpack plugins
    */
   plugins: [],
-  routes: {
-    '/api/provisioning': { host: `http://localhost:8000` }
-  },
+  routesPath: process.env.ROUTES_PATH && resolve(process.env.ROUTES_PATH),
   moduleFederation: {
     exposes: {
       './RootApp': resolve(__dirname, './src/AppEntry'),
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "lint:js:fix": "eslint config src --fix",
     "lint:sass": "stylelint 'src/**/*.scss' --config .stylelintrc.json",
     "patch:hosts": "fec patch-etc-hosts",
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -18,3 +18,4 @@ yarn-error.log*
 
 .DS_Store
 coverage
+.cache/*
```
ezr-ondrej, line 21: Can you add the newline here pls? :)

package.json
```diff
@@ -14,7 +14,8 @@
     "lint:js:fix": "eslint config src --fix",
     "lint:sass": "stylelint 'src/**/*.scss' --config .stylelintrc.json",
     "patch:hosts": "fec patch-etc-hosts",
-    "start": "fec dev",
+    "start": "PROXY=true ROUTES_PATH=/routes.config.js fec dev",
```
ezr-ondrej, line 17: Moving to config dir will help here as well, it looks weird to start relative path with `/`
```suggestion
    "start": "PROXY=true ROUTES_PATH=config/routes.config.js fec dev",
```

fec.config.js
```diff
@@ -14,9 +14,7 @@ module.exports = {
    * Add additional webpack plugins
    */
   plugins: [],
-  routes: {
-    '/api/provisioning': { host: `http://localhost:8000` }
-  },
+  routesPath: process.env.ROUTES_PATH && __dirname + process.env.ROUTES_PATH,
```
ezr-ondrej, line 17: We already have resolve available here, I think we can use it :)
```suggestion
  routesPath: process.env.ROUTES_PATH && resolve(process.env.ROUTES_PATH),
```

ezr-ondrej (commented):
Can we move the routes file under config so we do not pollute the root dir pls?

ezr-ondrej (commented):
Can we add it to a README pls? :)

ezr-ondrej (approved):
We can add it in a follow up. Thanks @amirfefer :+1:

PR #156, opened 2022-11-22
feat(progress): add reservation id as hidden field

ezr-ondrej:
In order to test the UI or connect to the progressing reservation for any other purposes, we need to expose the reservation id in the UI.

Fixes HMSPROV-359


ezr-ondrej:
@akhil-jha would this suffice? This adds the input only after the reservation id is acquired, but we might want to show it always only with empty value, also what identifier would you prefer for it to have?

ezr-ondrej:
@akhil-jha finally figured out what's happening, we were unsetting the reservationID to stop polling, which was straight forward, but not the best strategy. I've tried to remedy, could you try if you see the reservation_id now?

akhil-jha (approved):
Works! Thanks

PR #158, opened 2022-11-23
fix(HMSPROV-351) - use same onClose in wizard steps

amirfefer:


ezr-ondrej (approved):
I don't know how it gets passed, but it works as expected :shrug:
Thanks @amirfefer ! :+1:

PR #159, opened 2022-11-24
refs(HMSPROV-351) - hide confirm modal after a launch success

amirfefer:


ezr-ondrej (commented):
Would the wizard context be more appropriate to store this? It might be useful everywhere to know whether we are done. We might even want to store it as launchStatus and update the progress of launch there, so other components can change behaviour. But I might definitely be wrong, it's just my gut feeling :)
WDYT?

amirfefer:
the wizard context is not accessible on the top level, we can split the wizard root component to make it work, but I guess it makes sense to do so only if we will need the launch status in another step.

ezr-ondrej:
Ok, sure, let's not optimize prematurely and go with simple solution for now 👍

ezr-ondrej (approved):
Thanks @amirfefer ! 👍

PR #162, opened 2022-11-29
feat(HMSPROV-306) - add multi region select

amirfefer:
https://user-images.githubusercontent.com/11807069/204598061-23fce5aa-48d8-4c72-a3a9-ac30dd67b87d.mov

### DIFF
--- a/src/API/helpers.js
+++ b/src/API/helpers.js
@@ -1,1 +1,2 @@
 export const provisioningUrl = (path) => `/api/provisioning/v1/${path}`;
+export const imageBuilderURL = (path) => `/api/image-builder/v1/${path}`;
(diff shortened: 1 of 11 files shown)


ezr-ondrej:
The select works well, but unfortunatelly, it doesn't work in the backend, because of image-builder.
The clones doesn't seem to be available through `compose/{id}`, but rather `clones/{id}`, so we fail to launch in those regions. I'd like to hold off merging this until:

1. It works in the backend - I'll try to talk to @crossaine about how to approach this
2. We have QE coverage, or at least QE ready to cover this - might be nice to help QE on this :)

amirfefer:
@adiabramovitch is it the issue with the worker?  I don't think it is a blocker for this one, no? 🤔

adiabramovitch (approved):

PR #164, opened 2022-12-01
chore: add changelog and generator

lzap:
I propose to start tagging "releases" in the same cadence as we do for backend - every sprint. So I created 0.11 (roughly to the sprint start) and 0.12 for today (demo time). And copied the changelog generation script and generated changelog. I think a copy of the script will do, we should not change it too often.

If this is approved and merged, I will also push the tags.

### DIFF
--- a/scripts/changelog.py
+++ b/scripts/changelog.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+import os.path
+from pathlib import Path
+from typing import List, Optional
+
+from git_changelog import templates
+from git_changelog.build import Changelog
+from git_changelog.providers import GitHub, RefDef, RefRe
+
+JIRA_URL = "https://issues.redhat.com"
+JIRA_PROJECT = "HMSPROV"
+JIRA_ISSUE_SYMBOL = "HMSPROV-"
+
+GitHub.REF["issues"] = RefDef(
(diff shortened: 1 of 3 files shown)


scripts/changelog.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+import os.path
+from pathlib import Path
+from typing import List, Optional
+
+from git_changelog import templates
```
mshriver, line 9: not from the built-ins, time to introduce a requirements.txt in the scripts dir?

lzap: Yeah, actually this is a copy from https://github.com/RHEnVision/provisioning-backend/blob/main/scripts/changelog.py which is lazy. Perhaps we could create a new "pip module" and use it that way? I am not sure about terminology, I don't do much Python.

@ezr-ondrej creating a new git repo under the org and releasing it on pip should not be so hard I guess

ezr-ondrej:
Changelog is published in GH repo only for now, but already used in GH actions, happy to merge the makefile and changelog tho :)

lzap:
That's fine I will close there's not much. Pushing tags so we have a good starting point for the changelog for this sprint.

```
$ g push upstream --tags --no-verify
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
To https://github.com/RHEnVision/provisioning-frontend
 * [new tag]         0.11.0 -> 0.11.0
 * [new tag]         0.12.0 -> 0.12.0
```

PR #165, opened 2022-12-01
fix(HMSPROV-378) - use userEvent over fireEvent

amirfefer:
`userEvent` is made for mimicking  real user interactions with the screen, react testing library recommends using it  for testing user interactions (i.e typing, clicking, mouse focus)  over the lower api `fireEvent`


ezr-ondrej (approved):
Nice, the closer to user interaction we can get the better! :)

PR #171, opened 2022-12-27
Add gcp and azure to wizard

adiabramovitch:


ezr-ondrej:
Needs rebase :)

lzap:
Do we want to gate this via flags so we can hide everything but AWS in prod? That would mean integrating Unleash into the UI: https://docs.getunleash.io/reference/sdks/node

ezr-ondrej:
there's also an internal documentation for how to do this. If we get close to merge, we should definitely gate this by flags :)

PR #175, opened 2023-01-13
feat(HMSPROV-419): add a timeout session for reservation polling 

amirfefer:
This demonstrates another implementation for a timeout session with no use of additional timers. 
In addition, this PR reduces the number of round trips to the server, due to an interval that increased (not linear) for each try.

### DIFF
--- a/src/Components/ProvisioningWizard/steps/FinishProgress/index.js
+++ b/src/Components/ProvisioningWizard/steps/FinishProgress/index.js
@@ -15,20 +15,8 @@ import { useWizardContext } from '../../../Common/WizardContext';
 import { useMutation, useQuery } from 'react-query';
 import { createAWSReservation, createNewPublicKey, fetchAWSReservation } from '../../../../API';
 import './styles.scss';
-
-const pf_success_color_100 = '#3E8635';
-const pf_danger_color_100 = '#C9190B';
-const RESERVATION_POLLING_INTERVAL = 500;
-
-const steps = [
-  { description: 'Uploading SSH public key', progress: 0 },
-  { description: 'Creating AWS reservation', progress: 20 },
-  {
-    description: 'Waiting for AWS',
-    progress: 40,
-  },
-  { description: 'Launch is completed', progress: 100 },
-];
+import useInterval from '../../../Common/Hooks/useInterval';
+import { AWS_STEPS, PF_DANGER_100, PF_SUCCESS_100, POLLING_BACKOFF_INTERVAL } from './constants';
 
 const FinishStep = ({ imageID, setLaunchSuccess }) => {
   const [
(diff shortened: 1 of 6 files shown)


src/Components/ProvisioningWizard/steps/FinishProgress/index.js
```diff
@@ -60,7 +63,14 @@ const FinishStep = ({ imageID, setLaunchSuccess }) => {
 
   const { data: polledReservation } = useQuery(['reservation', reservationID], () => fetchAWSReservation(reservationID), {
     enabled: !!reservationID && activeStep < steps.length - 1 && !awsReservationError && !pubkeyError,
-    refetchInterval: RESERVATION_POLLING_INTERVAL,
+    refetchInterval: (data) => {
+      if (data?.success || !!data?.error) return false;
+      console.log(currentInterval);
```
ezr-ondrej, line 68: Forgotten debug? :)

src/Components/ProvisioningWizard/steps/FinishProgress/index.js
```diff
@@ -128,11 +140,14 @@ const FinishStep = ({ imageID, setLaunchSuccess }) => {
               <br />
               {polledReservation?.status}
               <span className="status-error">
+                {isSessionTimeout && 'Session timeout'}
```
ezr-ondrej, line 143: I'd try to include bit more info, so users know what really is happening.
```suggestion
                {isSessionTimeout && 'Session timed out, the reservation took too long to fulfill'}
```

ezr-ondrej (commented):
I like this approach quite a lot! :orange_heart:

src/Components/ProvisioningWizard/steps/FinishProgress/index.js
```diff
@@ -15,10 +15,11 @@ import { useWizardContext } from '../../../Common/WizardContext';
 import { useMutation, useQuery } from 'react-query';
 import { createAWSReservation, createNewPublicKey, fetchAWSReservation } from '../../../../API';
 import './styles.scss';
+import useInterval from '../../../Common/Hooks/useInterval';
 
 const pf_success_color_100 = '#3E8635';
 const pf_danger_color_100 = '#C9190B';
-const RESERVATION_POLLING_INTERVAL = 500;
+const POLLING_INTERVAL = [500, 1000, 1500, 2500, 4000, 6500, 10500, 17000];
```
lzap, line 22: I am curious where did you get these numbers? Assuming these are milliseconds, 17 seconds feels like too much for what it's doing (reservation is just query in the DB which us sub 100ms request). I suggest to keep the backoff interval in 500ms - 5s or something like that.

amirfefer: > I am curious where did you get these numbers? 

The numbers here are just sort of Fibonacci for the demonstration (yet, it's still POC :))

The main idea here is to make the backoff strategy dynamic with just a constant array, I think adding 100ms every interval is a bit naive and might add redundant roundtrips with no benefits to the user experience. we can use a mixed backoff strategy, which starts with a linear increase but ends with exponential backoff, i.e

500
600
700
800
900
1000
1250
1500
1750
2000
4000
8000
16000
32000
64000

ezr-ondrej: I'd agree with @amirfefer here, the initial is good to be linear, but then we can save some roundtrips (even tho those are fast, probably not necessary). I'd probably cut early (before the minute of wait (or add more frequent at the end) as the last two are quite large, but I guess that we should just try somthing off and if that does not work, then tweak it.

ezr-ondrej: One thought I had here was having some kind of identifier so we can track these requests belonging together. Maybe some additional header that we can track in our logger? Or do we think the `reservation_id` is enough?

src/Components/Common/Hooks/useTimeout.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+const useTimeout = (callback, delay) => {
```
ezr-ondrej, line 3: Can we remove `useTimeout` pls?

ezr-ondrej (approved):
Thanks @amirfefer ! :+1:

PR #176, opened 2023-01-13
fix(HMSPROV-424): hide unsupported types

amirfefer:
filter out instance types that aren't supported


ezr-ondrej:
I guess our CI expets some types to be there and those are not anymore. Are we going to loose ability to test with these instances?
If so should we hide this only if the feature flag is switched to Production? We could switch it on for Development so we could continue testing with as small instances as possible :thinking:

amirfefer:
clsong this for adding warning alert instead

PR #180, opened 2023-01-18
fix(HMSPROV-406): filter instance types with substring

amirfefer:
This PR allows filtering instance's types with any substring within the type, i.e -  for searching  `micro`, I'll get all the micro types.


lzap (approved):
Thank you, love it.

PR #188, opened 2023-01-29
feat(HMSPROV-1133): add instances description after a successful launch

amirfefer:
This PR shows instances list after a succesfull laucnh

https://user-images.githubusercontent.com/11807069/215360024-2c8a489c-9789-468b-852d-27ebf8ce75f1.mov

### DIFF
--- a/src/API/index.js
+++ b/src/API/index.js
@@ -33,3 +33,8 @@ export const fetchImageClones = async (composeID) => {
   const { data } = await axios.get(imageBuilderURL(`composes/${composeID}/clones`));
   return data;
 };
+
+export const fetchAWSInstances = async (reservationID) => {
+  const { data } = await axios.get(provisioningUrl(`reservations/aws/${reservationID}/instances`));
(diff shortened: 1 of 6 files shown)


lzap:
WFM, this should be a temporary solution until we integrate more with the inventory.

Small suggestion: These are called "Public IP" and "Public DNS" because there are also private IPs or DNS entries for instances on AWS. Maybe to reword this, no big deal.

lzap:
Are you going to continue on this one, or refactor this into what we talked about - backend doing the 5 second (max) wait and then providing the data in the reservation itself. So no additional refreshing is needed. We should cover 90% of all launches in a nice UX way.

amirfefer:
> Are you going to continue on this one, or refactor this into what we talked about 

as we dicussed, i'm going to refacor the backend part and then consuming the instance data direcctly from the reservation

amirfefer:
closed in favoe of #213

PR #193, opened 2023-02-07
refactor: Add custom ids to wizard buttons

ezr-ondrej:
Add custom ids to Provisioning wizard.
The buttons needed a CustomHeader for the id to be set.

Fixes HMS-1127

### DIFF
--- a/src/Components/ProvisioningWizard/CustomFooter.js
+++ b/src/Components/ProvisioningWizard/CustomFooter.js
@@ -1,0 +1,48 @@
+import * as React from 'react';
+import { Button, ButtonVariant, WizardFooter, WizardContextConsumer } from '@patternfly/react-core';
+import styles from '@patternfly/react-styles/css/components/Wizard/wizard';
+
+import { stepIdToString } from './steps';
+
+const CustomFooter = () => (
+  <WizardFooter>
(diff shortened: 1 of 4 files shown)


amirfefer (approved):
Thanks @ezr-ondrej 👍 

Tested manually and it works as expected, the ids are set correctly, 
Just wondering if we would like to add a custom id for monitoring thr `Edit` button, which navigates the user to the form.

akhil-jha:
> Tested manually and it works as expected, the ids are set correctly, Just wondering if we would like to add a custom id for monitoring thr `Edit` button, which navigates the user back to the form.

I guess only when it's a failed reservation the user will click on Edit, I don't know any other way to bring that edit modal on screen. If we are already collecting failed reservation, I think Edit would be a duplicate.

ezr-ondrej:
> I guess only when it's a failed reservation the user will click on Edit, I don't know any other way to bring that edit modal on screen. If we are already collecting failed reservation, I think Edit would be a duplicate.

I tend to agree, but can't hurt, I'm not sure how pendo actually works, so better to have more info, then less :+1:

ezr-ondrej:
Merging as the nit was added, so I think it was an ack from Amir :)

PR #194, opened 2023-02-13
refactor: Add sources filtering by provider

adiabramovitch:
This is a temporary fix until @amirfefer will finish adding the GCP and Azure UI part.

### DIFF
--- a/src/Components/DescriptionListAWS/index.js
+++ b/src/Components/DescriptionListAWS/index.js
@@ -7,9 +7,10 @@ import { SOURCES_QUERY_KEY } from '../../API/queryKeys';
 import { fetchSourcesList } from '../../API';
 import { useWizardContext } from '../Common/WizardContext';
 
+const PROVIDER = 'aws';
 const DescriptionListAWS = ({ imageName }) => {
   const [wizardContext] = useWizardContext();
-  const { error, data: sources } = useQuery(SOURCES_QUERY_KEY, fetchSourcesList);
+  const { error, data: sources } = useQuery(SOURCES_QUERY_KEY, () => fetchSourcesList(PROVIDER));
 
   if (error) {
     // TODO: error handling, notifications
--- a/src/Components/SourcesSelect/index.js
+++ b/src/Components/SourcesSelect/index.js
@@ -7,6 +7,7 @@ import { SOURCES_QUERY_KEY } from '../../API/queryKeys';
 import { fetchSourcesList } from '../../API';
 import { useWizardContext } from '../Common/WizardContext';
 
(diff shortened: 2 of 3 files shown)


src/Components/DescriptionListAWS/index.js
```diff
@@ -7,9 +7,10 @@ import { SOURCES_QUERY_KEY } from '../../API/queryKeys';
 import { fetchSourcesList } from '../../API';
 import { useWizardContext } from '../Common/WizardContext';
 
+const PROVIDER = 'aws';
 const DescriptionListAWS = ({ imageName }) => {
   const [wizardContext] = useWizardContext();
-  const { error, data: sources } = useQuery(SOURCES_QUERY_KEY, fetchSourcesList);
+  const { error, data: sources } = useQuery(SOURCES_QUERY_KEY, fetchSourcesList(PROVIDER));
```
ezr-ondrej, line 13: I don't think we necessarily need to filter here :)

src/Components/SourcesSelect/index.js
```diff
@@ -20,7 +21,7 @@ const SourcesSelect = ({ setValidation }) => {
     error,
     isLoading,
     data: sources,
-  } = useQuery(SOURCES_QUERY_KEY, fetchSourcesList, {
+  } = useQuery(SOURCES_QUERY_KEY, fetchSourcesList(PROVIDER), {
```
ezr-ondrej, line 24: we need to pass a function, not it's result here
```suggestion
  } = useQuery(SOURCES_QUERY_KEY, () => fetchSourcesList(PROVIDER), {
```
Or another solution would be to change `fetchSourcesList` to return a function

ezr-ondrej (changes requested):

src/Components/DescriptionListAWS/index.js
```diff
@@ -7,9 +7,10 @@ import { SOURCES_QUERY_KEY } from '../../API/queryKeys';
 import { fetchSourcesList } from '../../API';
 import { useWizardContext } from '../Common/WizardContext';
 
+const PROVIDER = 'aws';
```
amirfefer, line 10: with  #195 the `provider` is given by IB and stored in the wizard context

amirfefer (commented):
Thanks @adiabramovitch 👍 
a rebase with #195 will  give you the provider with no extra constants

In second thought, I've refacored the final descriptation page for general provider, so rebased would be quicker doing the rebase in #195 

Tested with aws and gce sources, works as expected

amirfefer (approved):

PR #195, opened 2023-02-13
feat(hms-776): add gcp UI to wizard

amirfefer:
This PR adds google cloud provisioning to the wizard:
Screen Shot 2023-02-09 at 22 40 39


amirfefer:
all 🟢  :)

lzap:
My JS reading skills are so bad so I have to ask: what the two steps do exactly? Is that the first is create reservation call and the second is waiting until it is finished? In that case I suggest wording it accordingly, something like:

Create reservation
Submit requested data

Launch instance(s)
Call Google API

I think we should use the same wording for all hyperscalers.

What happens when we are creating a new SSH pubkey, is this done when user click in the wizard or should that be an extra step here? (Existing key that is fine no need to upload it to our database).

Other than that, well, looks great.

akhil-jha:
Just reiterating, pr_check is not testing GCP as there is no test for it. :)

ezr-ondrej:
Do we have accompanying PR to image builder for passing the provider? I'm bit worried what happens when we merge and the PR is not in.
It has conflicts with the two small PR merged recently

amirfefer:
> Do we have accompanying PR to image builder for passing the provider? @ezr-ondrej 

Yes, It's mentioned above - https://github.com/RedHatInsights/image-builder-frontend/pull/965

> What happens when we are creating a new SSH pubkey, is this done when user click in the wizard or should that be an extra step here? (Existing key that is fine no need to upload it to our database). @lzap 

Same behaivier with AWS, extra step  - 
Screen Shot 2023-02-14 at 15 08 58

>  what the two steps do exactly? Is that the first is create reservation call and the second is waiting until it is finished?

Iindeed, the `create reservation` reflects the form / data submission to our service
the  `Launch instance(s)` is the backend's job, which has 1 step.
I agree with the terminology

amirfefer:
Rebased and updated

src/API/index.js
```diff
@@ -11,13 +11,13 @@ export const fetchPubkeysList = async () => {
...
-export const fetchInstanceTypesList = async (region) => {
-  const { data } = await axios.get(provisioningUrl(`instance_types/aws?region=${region}`));
+export const fetchInstanceTypesList = async (region, provider) => {
+  const { data } = await axios.get(provisioningUrl(`instance_types/${provider}?region=${region}`));
   return data;
 };
 
-export const createAWSReservation = async (params) => {
-  return axios.post(provisioningUrl('reservations/aws'), params);
+export const createReservation = (provider) => async (params) => {
```
ezr-ondrej, line 19: could we keep the functions here consistent? I know this is bit better, but in other cases we need parameters, we do `() => createReservation(provider)` on the caller (in the `useQuery`).

src/Components/ProvisioningWizard/steps/AccountCustomizations/index.js
```diff
@@ -0,0 +1,39 @@
...
+
+  React.useEffect(() => {
+    provider &&
+      setWizardContext((prevState) => ({
+        ...prevState,
+        provider,
+        chosenRegion: defaultRegionByProvider(provider),
+        chosenImageID: composeID,
+      }));
+  }, [provider, setWizardContext]);
```
ezr-ondrej, line 20: This feels bit out of place, shouldn't we rather set the provider in the parent component (`ProvisioningWizard`) and only use it from the context here?

ezr-ondrej (approved):
Minor nits, lets merge and let me know if you want to address the nits in follow up or if I should take a look into it :)

PR #196, opened 2023-02-14
fix(HMS-1237): show warning alert when instance type is not supported

amirfefer:
Screen Shot 2023-02-14 at 17 48 14

### DIFF
--- a/src/Components/InstanceTypesSelect/index.js
+++ b/src/Components/InstanceTypesSelect/index.js
@@ -14,6 +14,7 @@ const InstanceTypesSelect = ({ setValidation, architecture }) => {
   const [numOptions, setNumOptions] = React.useState(OPTIONS_PER_SCREEN);
   const [filteredTypes, setFilteredTypes] = React.useState(null);
   const [prevSearch, setPrevSearch] = React.useState('');
+  const [isTypeSupported, setTypeSupported] = React.useState(true);
   const {
     isLoading,
     error,
@@ -43,6 +44,8 @@ const InstanceTypesSelect = ({ setValidation, architecture }) => {
     if (isPlaceholder) {
       clearSelection();
     } else {
+      const chosenInstanceType = instanceTypes.find((instanceType) => selection === instanceType.name);
+      setTypeSupported(chosenInstanceType.supported);
       setWizardContext((prevState) => ({
         ...prevState,
         chosenInstanceType: selection,
@@ -54,6 +57,7 @@ const InstanceTypesSelect = ({ setValidation, architecture }) => {
 
   const clearSelection = () => {
     setValidation('error');
+    setTypeSupported(true);
     setWizardContext((prevState) => ({
       ...prevState,
       chosenInstanceType: null,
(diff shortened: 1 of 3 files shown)


lzap:
I am currently working on a backend feature that will fail the launch if such image is selected. I like that, we do not confuse users about missing types, it is pretty clear what is going on. Maybe rewording it a little to indicate what will happen if they continue:

The selected specification does not meet minimum requirements for this image. Launch will fail.

Not sure, also check with UX if we should add "Warning:" prefix or not. But works for me!

src/Components/InstanceTypesSelect/index.js
```diff
@@ -106,26 +110,37 @@ const InstanceTypesSelect = ({ setValidation, architecture }) => {
...
-      {selectItemsMapper(types, numOptions)}
-    </Select>
+    <>
+      {!isTypeSupport && (
+        <Alert
+          data-testid="unsupported_type_alert"
+          ouiaId="instance_type_not_supported_alert"
+          variant="warning"
+          isInline
+          title="Warning: The selected type is not sufficient for the chosen image"
```
ezr-ondrej, line 120:
```suggestion
          title="Warning: The selected type does not  have sufficient resources for the chosen image"
```

amirfefer: @lzap suggested the following warning:
`The selected specification does not meet minimum requirements for this image. Launch will fail.`

MariSvirik:
Do we even want to allow users to select an unsupported type and proceed to the next step? I assume we know right away which types are supported, right?

**1.option:** disable unsupported types - we can add helper text to tell users that this is not supported. Users see them in the dropdown but cannot select them.

**2. option:**  If we want to allow selection
I would go with an error helper text under the field and the icon within the field. But I would keep ‘next’ disabled. See [PF4 guidelines here.](https://www.patternfly.org/v4/components/form/design-guidelines#errors-and-validation)


Screenshot 2023-02-15 at 12 32 12

Or am I missing something?

amirfefer:
Thanks @MariSvirik, I'd pick option 2, though currently  we use this warning alert  in other inputs, I guess it better to refacor all inputs all together to use PF4 inline validations for warnings / errors in a follow up PR.

src/Components/InstanceTypesSelect/index.js
lzap: All I drop into comments are just ideas, make your own pick. Or if you don’t like it, that is fine.

Maybe "Launch will fail" is too strict, so only the first sentence? Or I like the fixed version (@ezr-ondrej proposal) too.

lzap:
We talked about this and we believe we should not hide well-known types to experienced EC2 users, we want to have a warning.

Now, we already have warnings and errors implemented in one way. Let’s stick with this as this is a high priority PR. @amirfefer will create an issue to refactor all our warnings/errors into the recommended PF4 components.

ezr-ondrej (approved):
oki, lets do it :+1: 

I'd love to see the refactoring of the UI soon now tho.
Error states are not as likely to be exposed, but this will be exposed to many users.

ezr-ondrej:
Needs rebase tho

ezr-ondrej (approved):
Thanks @amirfefer :+1:

PR #199, opened 2023-02-16
build: remove node_modules from SonarQube scan

ezr-ondrej:
Even ignored it is noisy, lets nuke 'em


mshriver (approved):
Hopefully the hammer doesn't get any larger.

ezr-ondrej:
Lets not do this, if we really need to, we can push the sonar as first action, before `npm install`

PR #200, opened 2023-02-20
chore: fix default statement issues

ezr-ondrej:
It is considered a security risk.


ezr-ondrej:
I'll merge as this is not the cause and I need to fix the sonar findings :)

PR #201, opened 2023-02-20
build: include coverage in sonar scan

ezr-ondrej:


ezr-ondrej:
The sonarqube passed.

PR #209, opened 2023-03-01
fix(HMS-1131): filter out regions with unsuccessful image clones

amirfefer:
This PR filter out regions that the clone process ended unsuccessful

### DIFF
--- a/src/API/index.js
+++ b/src/API/index.js
@@ -34,6 +34,11 @@ export const fetchImageClones = async (composeID) => {
   return data;
 };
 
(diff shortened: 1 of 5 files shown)


ezr-ondrej (approved):
We are not testing for the pending status, but manually confirmed this works well and the most important is the failed, great job here, love it :+1:

Ofc it would be great to talk to image builder team to include the status in the list as here we can already notice the lag in the load, but this is the best we can do for now, thanks @amirfefer ! :+1:

PR #210, opened 2023-03-01
fix(HMS-1361): fix a 404 api call at first rendering

amirfefer:
the `provider` is undefined at first rendering,  causes an extra triggering of `/sources?provider=undefiend` which ended up with 404


ezr-ondrej (approved):
Thanks @amirfefer ! :+1:

PR #222, opened 2023-03-12
Add support to a source shared with an image

adiabramovitch:
This PR matches Image builder change to share images with a source rather than an account ID (AWS for now). 
Image builder have added a prop name: sourceId on their side. (currently, image builder UI supports only sharing with one source and the provisioning service as well)
In this PR we are preselecting the source to the one given by IB, and showing it as disabled (You cannot change the source from the dropdown). 
This addition currently is only for AWS, in GCP and Azure the source should be selected from the dropdown. 

_Good To Know:_ 
After the change IB did it is **not possible** to use sources locally, you should use sources stage, make sure you have added a source there. 
If the image was not shared with a source the Launch button is not going to appear. (in AWS we do not allow (for now?) an image to be shared with an account id)
image

### DIFF
--- a/src/Components/SourcesSelect/SourcesSelect.test.js
+++ b/src/Components/SourcesSelect/SourcesSelect.test.js
@@ -4,10 +4,22 @@ import userEvent from '@testing-library/user-event';
 import { sourcesList } from '../../mocks/fixtures/sources.fixtures';
 import { render, screen } from '../../mocks/utils';
 
+const mockContext = require('../Common/WizardContext/initialState');
+
 describe('SourcesSelect', () => {
-  test('preselects the chosen source', async () => {
-    render(<SourcesSelect setValidation={jest.fn()} />);
+  test('preselect aws source', async () => {
+    render(<SourcesSelect imageSourceID={1} setValidation={jest.fn()} />);
     // wizard context mock is `{chosenSource: 1}`
+    const selectDropdown = await screen.findByText('Source 1');
+    await userEvent.click(selectDropdown);
+  });
+
+  test('preselects the chosen source', async () => {
+    jest.mock('../Common/WizardContext/initialState', () => ({
+      __esModule: true,
+      default: { ...mockContext, provider: 'gcp' },
+    }));
+    render(<SourcesSelect setValidation={jest.fn()} />);
     const selectDropdown = await screen.findByText('Source 1');
     await userEvent.click(selectDropdown);
 
--- a/src/Components/SourcesSelect/index.js
+++ b/src/Components/SourcesSelect/index.js
@@ -2,12 +2,12 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { Alert, Select, SelectOption, Spinner } from '@patternfly/react-core';
 import { useQuery } from 'react-query';
-
 import { SOURCES_QUERY_KEY } from '../../API/queryKeys';
 import { fetchSourcesList } from '../../API';
 import { useWizardContext } from '../Common/WizardContext';
+import { IB_SOURCE_PROVIDERS } from '../Common/constants';
 
-const SourcesSelect = ({ setValidation }) => {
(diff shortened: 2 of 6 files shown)


src/Components/SourcesSelect/index.js
```diff
@@ -24,9 +24,18 @@ const SourcesSelect = ({ setValidation }) => {
     enabled: !!provider,
     onSuccess: (data) => {
       const id = chosenSource;
-
-      if (!id) return;
-      setSelected(selectObject(id, data.find((source) => source.id === id).name));
+      if (provider === AWS_PROVIDER) {
+        setSelected(selectObject(imageSourceID, data.find((source) => source.id === imageSourceID).name));
```
ezr-ondrej, line 28: The error here is that source.id (comming from the mocked response) is an integer, but the imageSourceID is a string. I guess `==` would fix this as it would allow type mismatch :)

ezr-ondrej (commented):
Hi! :) I've left few nits, but a general question: 

Do we actually want to preselect the Source, or disable it all together?

I'd just disable the select, when we have the ID defined.
This would make us do two tests one with the ID passed and second without, just checking we have the input disabled.

src/Components/SourcesSelect/index.js
```diff
@@ -69,8 +78,9 @@ const SourcesSelect = ({ setValidation }) => {
       isOpen={isOpen}
       onToggle={(openState) => setIsOpen(openState)}
       selections={selected}
+      isDisabled={!!imageSourceID}
       onSelect={onSelect}
-      placeholderText="Select account"
+      placeholderText={'Select account'}
```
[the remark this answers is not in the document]
adiabramovitch: not needed, don't remember changing it maybe the linter? 
replaced it to `"Select account"`

src/Components/SourcesSelect/index.js
adiabramovitch: didn't help :(

src/Components/SourcesSelect/index.js
```diff
@@ -24,9 +24,18 @@ const SourcesSelect = ({ setValidation }) => {
     enabled: !!provider,
     onSuccess: (data) => {
       const id = chosenSource;
-
-      if (!id) return;
-      setSelected(selectObject(id, data.find((source) => source.id === id).name));
+      if (provider == AWS_PROVIDER) {
```
ezr-ondrej, line 27: should we limit this for only the cases the source ID was passed?
```suggestion
      if (provider == AWS_PROVIDER && imageSourceID) {
```
That way we would allow for the provisioning of images shared with account.

1. it would allow to keep the simple pipeline we have currently (testing is easier)
2. if we ever decide to allow for the account shared images, we can do so easily :)

adiabramovitch: sure, I don't have a problem just thought we are not enabling it. 
But maybe it will be smarter to start this way and change in the future if we see fit.

src/Components/SourcesSelect/SourcesSelect.test.js
```diff
@@ -5,13 +5,23 @@ import { sourcesList } from '../../mocks/fixtures/sources.fixtures';
...
 describe('SourcesSelect', () => {
+  test('preselect aws source', async () => {
+    render(<SourcesSelect imageSourceID={1} setValidation={jest.fn()} />);
+    // wizard context mock is `{chosenSource: 1}`
+    const selectDropdown = await screen.findByText('Source 1');
+    await userEvent.click(selectDropdown);
+  });
+
   test('preselects the chosen source', async () => {
+    const mockContext = require('../Common/WizardContext/initialState');
```
amirfefer, line 16: please move importing to the top level

amirfefer (commented):
Thanks @adiabramovitch !

ezr-ondrej (approved):
Thank you @adiabramovitch, works well :+1:
Thanks @amirfefer for review :+1:

PR #229, opened 2023-03-23
fix(HMS-#1504): fix timeout interval regression during lanuch

amirfefer:
Timeout error didn't appear correctly during launch progress, casued by a regression after switching to the progress steps. a test is added  for preventing such regression

### DIFF
--- a/src/Components/Common/Hooks/useInterval.js
+++ b/src/Components/Common/Hooks/useInterval.js
@@ -15,7 +15,7 @@ const useInterval = (list) => {
 
   const nextInterval = () => {
     const nextValue = generatorRef.current.next().value;
-    setCurrentInterval(nextValue);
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Thanks @amirfefer ! :+1: 
Great fix, happy to see the test.

Note: it got me thinking about messaging, we should make sure Users clearly know we've stopped polling for results, but didn't stop the reservation.
Might be worth thinking about what we want to do in those cases.
CC @lzap

lzap:
Yeah I think I am thinking the same as Ondra. I think as part of every user message, there should be a resolution. In this case, "Check your AWS console. You may close this window." or something like that.

This is a general rule I am trying to live with, in the worst possible errors there should be at least "Contact Red Hat Support." if we know for fact it is bad and we can't recommend anything else :-)

PR #234, opened 2023-03-31
feat(HMS-1271): add popular group for instances types

amirfefer:
Screen Shot 2023-03-31 at 15 47 37

### DIFF
--- a/src/Components/InstanceTypesSelect/InstanceTypesSelect.test.js
+++ b/src/Components/InstanceTypesSelect/InstanceTypesSelect.test.js
@@ -1,9 +1,10 @@
 import React from 'react';
 import userEvent from '@testing-library/user-event';
 import InstanceTypesSelect from '.';
-import { awsInstanceTypeList, azureInstanceTypeList } from '../../mocks/fixtures/instanceTypes.fixtures';
+import { awsInstanceTypeList, AWSPopularType, AWSTypesWithPopularType, azureInstanceTypeList } from '../../mocks/fixtures/instanceTypes.fixtures';
 import { render, screen } from '../../mocks/utils';
 import initialWizardContext from '../Common/WizardContext/initialState';
+import { provisioningUrl } from '../../API/helpers';
 
 describe('InstanceTypesSelect', () => {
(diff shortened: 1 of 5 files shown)


lzap:
Looks great.

ezr-ondrej:
WDYT @MariSvirik ? :)

MariSvirik:
What is under the divider? All types or the rest?
I'm not sure if this is visually enough to distinguish.
But the group itself is fine.

lzap:
One UX idea: I would not show popular items in search results. Only when you open up the dropdown without typing anything. Because if we have too many items in the popular section we are doing it wrong :-)

lzap:
What is the status?

ezr-ondrej:
> How many popular items do we plan to have?

This is really up to us :) We need to decide on these, so if we could find ways to measure the success of this (maybe Pendo for how many users actually choose Popular instances) that'd be a great win IMO

lzap:
To me one popular item from three major categories would be great:

* General purpose instance (all-rounder).
* CPU-optimized category (intensive tasks).
* Memory intensive category (memory intensive tasks).

Although cloud providers have many other categories and offerings (network optimized, storage optimized, ephemeral storage, SSD NVMe super fast local storage) we would sink into the rabbit hole fairly quickly. So picking just three favourite  ones make sense for me.

adiabramovitch:
@amirfefer can fix the conflicts? I see the pipeline is fixed 🎊

amirfefer:
@adiabramovitch, I guess this is on hold until we have more user insights

lzap:
After chat on our todays planning, I have rewritten the Jira issue into a different story: _Implement a "bookmarking" feature for the instance type dropdown when users are able to select arbitrary amount of entries in the dropdown by clicking on some "star" icon or something like that. The favourite list is then presented at the very start of the listing so it is accessible. The data is stored in the user browser so it is available on each visit._

The reason for this was that we cannot agree on what makes a type a popular type, so let our users to decide themselves.

amirfefer:
Closing this for now, not enough data.

PR #236, opened 2023-04-03
feat: Prolong polling for Azure

ezr-ondrej:
Azure reservation tends to take longer then other providers. This allows to adjust the polling lenght per provider. For Azure we add 3 extra minutes of polling.

Refs HMS-1553


amirfefer (commented):
Thanks @ezr-ondrej ! I left a few comments

lzap:
I am okay, tho, I would vote for having the same wait time for every provider. It is not that big of a deal I think. We already have the "taking too long" message I think.

amirfefer:
@ezr-ondrej, what is the status here? would we want it before freeze?

ezr-ondrej:
> @ezr-ondrej, what is the status here? would we want it before freeze?

not sure, I don't think this is great solution if it can fail, it adds complexity... I'm thinking: lets add two minutes for all the providers for now and lets rethink the approach post summit?

I'd think we can maybe poll forever and just in few seconds show "this is taking bit too long, sorry, but you can close this"

ezr-ondrej:
I've decided just to prolong this all for one minute, lets figure out how this works in practice and then come up with a proper solution :)

lzap (approved):
I think this might be safe to be merged right now, up to you @amirfefer

PR #237, opened 2023-04-04
test: enable Azure wizard from Sample page

ezr-ondrej:
Screenshot from 2023-04-04 10-16-46

### DIFF
--- a/src/Routes/SamplePage/SamplePage.js
+++ b/src/Routes/SamplePage/SamplePage.js
@@ -1,14 +1,13 @@
 import React, { useEffect } from 'react';
 import { withRouter } from 'react-router-dom';
 
-import { Button, StackItem, Stack, Title, FormGroup, TextInput, Select, SelectOption, SelectVariant } from '@patternfly/react-core';
-import { Main } from '@redhat-cloud-services/frontend-components/Main';
+import { Button, GridItem, Grid, Title, FormGroup, TextInput, Select, SelectOption, SelectVariant } from '@patternfly/react-core';
 import { PageHeader, PageHeaderTitle } from '@redhat-cloud-services/frontend-components/PageHeader';
 import axios from 'axios';
 
 import './sample-page.scss';
 import ProvisioningWizard from '../../Components/ProvisioningWizard';
-import { AWS_PROVIDER } from '../../Components/Common/constants';
+import { AWS_PROVIDER, AZURE_PROVIDER } from '../../constants';
 
 /**
 This page demonstrates the provisioning UI wizard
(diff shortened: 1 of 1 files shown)


src/Routes/SamplePage/SamplePage.js
```diff
@@ -82,23 +86,36 @@ const SamplePage = () => {
...
-          </StackItem>
-        </Stack>
-      </Main>
+      <Grid hasGutter>
+        {images && <GridItem>{renderSelect(images)}</GridItem>}
+        <GridItem span={3}>
+          <FormGroup label="Provider">
+            <Select
+              variant={SelectVariant.single}
+              aria-label="Select Provider"
```
akhil-jha, line 95: Can you add this as well? :) 
```suggestion
              aria-label="Select Provider"
              ouiaId="select_provider"
```

src/Routes/SamplePage/SamplePage.js
```diff
@@ -82,23 +86,36 @@ const SamplePage = () => {
...
+              isOpen={isProviderSelectOpen}
+            >
+              <SelectOption key={AWS_PROVIDER} value={AWS_PROVIDER} />
+              <SelectOption key={AZURE_PROVIDER} value={AZURE_PROVIDER} />
+            </Select>
+          </FormGroup>
+        </GridItem>
+        <GridItem span={6}>
+          <FormGroup label="Image">
+            <TextInput id="ami" value={chosenImage.id || ''} onChange={onInputChange} />
```
akhil-jha, line 108:
```suggestion
            <TextInput id="ami" ouiaId="image_id"  value={chosenImage.id || ''} onChange={onInputChange} />
```

akhil-jha (changes requested):
Since we have a dropdown as well now, having ouids help in picking out what to select based on provider type in test :)

src/Routes/SamplePage/SamplePage.js
```diff
@@ -23,10 +22,11 @@ const SamplePage = () => {
   const [chosenImage, setChosenImage] = React.useState({
     id: undefined,
     name: undefined,
-    provider: undefined,
+    provider: AWS_PROVIDER,
```
akhil-jha, line 25: Does this need to have a value? What if we keep it to undefined?
I am asking because in testing, if the value is already selected, it won't perform the select operation and fail the test there. Workaround would be to add unnecessary if else block to see what is already selected 
But if nothing is selected as default,  that would work.

akhil-jha: Update: Nevermind. That still works.

PR #238, opened 2023-04-06
feat: add empty state when no source available

ezr-ondrej:
Adds empty state when no Source that can Launch given image is found. Provides link to Source creation.
Makes sure to refetch the empty state regularly so users do not need to refresh page. Adds fallback link for users who want to launch through cloud providers console directly.

Refs HMS-1583

NEW only empty state update - with region: 
[Screencast from 2023-04-13 14-02-44.webm](https://user-images.githubusercontent.com/2884324/231752847-2c5abab5-3ed3-43c7-a022-69d9ce0b62fb.webm)

~~Old: [Screencast from 2023-04-07 01-08-48.webm](https://user-images.githubusercontent.com/2884324/230512943-f3717717-9c06-4545-b08c-2975f60a4fee.webm)~~

- [x] ~Depends on~: https://github.com/RedHatInsights/image-builder-frontend/pull/1061

### DIFF
--- a/src/API/queryKeys.js
+++ b/src/API/queryKeys.js
@@ -1,4 +1,5 @@
-export const SOURCES_QUERY_KEY = 'sources';
+export const sourcesQueryKey = (provider) => ['sources', provider];
+export const sourceUploadInfoKey = (source_id) => ['source_upload_info', source_id];
 export const PUBKEYS_QUERY_KEY = 'pubkeys';
 export const instanceTypesQueryKeys = (region) => ['instanceTypes', region];
 export const IMAGE_REGIONS_KEY = 'image_region';
--- a/src/Components/ProvisioningWizard/index.js
+++ b/src/Components/ProvisioningWizard/index.js
@@ -6,7 +6,8 @@ import { WizardProvider } from '../Common/WizardContext';
 import APIProvider from '../Common/Query';
 import ConfirmModal from '../ConfirmModal';
 import CustomFooter from './CustomFooter';
-import defaultSteps from './steps';
+import { imageProps, sourcesForImage } from './helpers';
+import getSteps from './steps';
 import './steps/Pubkeys/pubkeys.scss';
 
 const DEFAULT_STEP_VALIDATION = {
@@ -19,6 +20,8 @@ const ProvisioningWizard = ({ isOpen, onClose, image, ...props }) => {
   const [stepValidation, setStepValidation] = React.useState(DEFAULT_STEP_VALIDATION);
   const [isConfirming, setConfirming] = React.useState(false);
   const [successfulLaunch, setLaunchSuccess] = React.useState();
+
+  const { isLoading, error: sourcesError, sources: availableSources } = sourcesForImage(image, { refetch: stepIdReached <= 2 });
 
   const onCustomClose = () => {
     setConfirming(false);
@@ -35,7 +38,10 @@ const ProvisioningWizard = ({ isOpen, onClose, image, ...props }) => {
     }
   };
 
-  const steps = defaultSteps({
+  const steps = getSteps({
+    sourcesError,
+    isLoading,
+    availableSources,
     stepIdReached,
     image,
     stepValidation,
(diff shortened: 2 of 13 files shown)


src/Components/ProvisioningWizard/steps/index.js
```diff
@@ -32,7 +42,7 @@ const defaultSteps = ({
             provider={provider}
             architecture={architecture || 'x86_64'}
             composeID={id}
-            imageSourceID={sourceId}
+            imageSourceID={sourceIDs?.[0]}
```
ezr-ondrej, line 45: This is temporary, to keep this small, follow up should be to get the Sources select to filter Sources the same way.

src/Components/ProvisioningWizard/index.js
```diff
@@ -20,6 +21,10 @@ const ProvisioningWizard = ({ isOpen, onClose, image, ...props }) => {
   const [isConfirming, setConfirming] = React.useState(false);
   const [successfulLaunch, setLaunchSuccess] = React.useState();
 
+  const { isLoading, error: sourcesError, sources: availableSources } = sourcesForImage(image, { refetch: stepIdReached <= 2 });
+
+  if (isLoading) return null;
```
ezr-ondrej, line 26: This is a shortcut, it would probably be better to open the empty stater right away with spinner.
Should get done before merge IMO

ezr-ondrej: Fixed now

adiabramovitch (approved):
LGTM 
let's get it in and do follow ups if needed 
@ezr-ondrej Thanks!

src/Components/ProvisioningWizard/index.js
```diff
@@ -48,35 +54,32 @@ const ProvisioningWizard = ({ isOpen, onClose, image, ...props }) => {
...
+        className={'provisioning'}
+        footer={<CustomFooter />}
+      />
+      <ConfirmModal isOpen={isConfirming} onConfirm={onCustomClose} onCancel={() => setConfirming(false)} />
     </WizardProvider>
-  ) : null;
+  );
 };
 
+const ProvisioningWizardWrapper = (props) => <APIProvider>{props.isOpen && <ProvisioningWizard {...props} />}</APIProvider>;
```
amirfefer, line 75: nit? Wouldn't it be better to spilt this into a container that cosumes the wizard component wrapped by the providers (context and API) ?

src/API/queryKeys.js
```diff
@@ -1,4 +1,5 @@
-export const SOURCES_QUERY_KEY = 'sources';
+export const sourcesQueryKey = (provider) => ['sources', provider];
+export const sourceUploadInfoKey = (source_id) => ['source_upload_info', source_id];
```
amirfefer, line 2: Could you please use the key directly and skip this helper function wrapper? these helper functions create a new array on every re-render.

The reason I stoped using it is also because `useQuery` changed the queries  keys, now arrays are mandatory for performence wise

src/Components/RegionsSelect/index.js
```diff
@@ -29,12 +29,17 @@ const RegionsSelect = ({ provider, currentRegion, composeID, onChange }) => {
   // filter successful clones images
   if (clonesStatusQueries.length && clonesStatusQueries.every((cloneQuery) => cloneQuery.isLoading === false)) {
     const clonesStatus = clonesStatusQueries?.map((query) => query?.data);
-    const filteredCloned = clonedImages?.filter((_, index) => clonesStatus[index].status === 'success');
-    images.push(...filteredCloned);
+    // enrich the cloned image data
+    clonesStatus?.forEach((status, index) => {
+      clonedImages[index] = { ...clonedImages[index], ...status?.options };
```
amirfefer, line 34: do we use the status options?

ezr-ondrej: Yes, in the direct link :)

ezr-ondrej:
Thanks for the review @amirfefer ! :orange_heart: 
I'll work on a follow-up that will use the "hooks" in the source select and improve them to follow the rules and address all your other comments

ezr-ondrej:
Follow-up: #246

PR #239, opened 2023-04-11
feat: Add Azure instance table on Launch success

ezr-ondrej:
Adds Instance table at the Launch success screen.

Adds instance link humanize function, as for Azure, we want shorten the visible ID.

Allows passing default values overrides for the WizardContext, this enables testing with different providers easily.

Refs HMS-1409

Screenshot from 2023-04-11 17-35-59


src/Components/Common/instanceHelpers.js
```diff
@@ -0,0 +1,21 @@
+import { AWS_PROVIDER, AZURE_PROVIDER } from './constants';
+
+export const instanceLink = (instanceID, provider, region) => {
+  switch (provider) {
+    case AWS_PROVIDER:
+      return `https://console.aws.amazon.com/ec2/home?region=${region}#InstanceDetails:instanceId=${instanceID}`;
+    case AZURE_PROVIDER:
+      return `https://portal.azure.com/#@rhdevcloudops.onmicrosoft.com/resource${instanceID}/overview`;
```
lzap, line 8: It reads `resourceXXXX` immediately without any separator, is that correct?

lzap: Also the tenant is hardcoded?

fec.config.js
```diff
@@ -2,7 +2,7 @@
 const { resolve } = require('path');
 
 module.exports = {
-  appUrl: '/apps/provisioning',
+  appUrl: '/insights/provisioning',
```
lzap, line 5: Is this intentional?

src/Components/Common/instanceHelpers.js
ezr-ondrej: Yeah, it is the Azure instanceID is of a format (I'm simplifying - but it starts with `/`) `/subscription/<sub_id>/resource_group/<resource_group_name>/Microsoft.Compute/<instance_name>`

lzap (approved):
Other than that...

src/Components/Common/instanceHelpers.js
lzap: Gotcha and our tenant is hardcoded because that is the service account I guess.

PR #240, opened 2023-04-11
refactor: Stateless region select

ezr-ondrej:
Make the region select stateless.

This allows for use of the select on multiple places.


lzap (approved):
Since tests are passing and we need to move on for the upcoming sources changes and Amir is not here, I will be merging few PRs here and there.

PR #244, opened 2023-04-13
feat(HMS-1109): Add correlation id to UI

adiabramovitch:


src/Components/ProvisioningWizard/index.js
```diff
@@ -20,6 +21,10 @@ const ProvisioningWizard = ({ isOpen, onClose, image, ...props }) => {
   const [isConfirming, setConfirming] = React.useState(false);
   const [successfulLaunch, setLaunchSuccess] = React.useState();
 
+  React.useEffect(() => {
+    axios.defaults.headers.common['Correlation-Id'] = uuidv4();
```
ezr-ondrej, line 25: The correlation ID starts to be bit overused, I'd go with something more specific here, smthing like:
```suggestion
    axios.defaults.headers.common['Provisioning-Wizard-Id'] = uuidv4();
```
@lzap opinions?

lzap: I don’t care which one to use, I already filed correlation-id for responses which is actually a trace id in our codebase. You know what, that is really weird, let me rename my header to Trace-Id and I propose this to be just Wizard-Id or maybe Ui-Id?

Anyways, you send a header but it will be thrown away by our backend. What is the purpose of this?

src/Components/SourcesSelect/index.js
```diff
@@ -26,7 +26,7 @@ const SourcesSelect = ({ setValidation, imageSourceID }) => {
       const id = chosenSource;
 
       if (IB_SOURCE_PROVIDERS.includes(provider) && imageSourceID) {
-        setSelected(selectObject(imageSourceID, data.find((source) => source.id === imageSourceID).name));
+        setSelected(selectObject(imageSourceID, data.find((source) => source.id == imageSourceID).name));
```
ezr-ondrej, line 29: FYI I'm removing this code in #246
If you'd not mind I guess it would be better to drop this and related changes, to reduce conflicts.

src/Components/ProvisioningWizard/index.js
ezr-ondrej: I guess this is a UI part of a change, this is supposed to be followed by backend PR if I get that correctly?
Would you mind covering this in tickets @adiabramovitch, so we have more clarity on it?

src/Components/ProvisioningWizard/index.js
```diff
@@ -20,9 +21,12 @@ const ProvisioningWizard = ({ isOpen, onClose, image, ...props }) => {
   const [stepValidation, setStepValidation] = React.useState(DEFAULT_STEP_VALIDATION);
   const [isConfirming, setConfirming] = React.useState(false);
   const [successfulLaunch, setLaunchSuccess] = React.useState();
-
   const { isLoading, error: sourcesError, sources: availableSources } = sourcesForImage(image, { refetch: stepIdReached <= 2 });
 
+  React.useEffect(() => {
+    axios.defaults.headers.common['Provisioning-Wizard-Id'] = uuidv4();
```
amirfefer, line 27: what about adding a cleanup when the wizard unmounts? just to be safe that we don't change axios globally. 
```js
useEffect(() => {
  performSideEffect();
  return cleanUpFunction;
}, []);
```
We need also to be carefull due to react 18, which `useEffect` triggers twice (so we might end up with two distinct ids on the same launch)

I'm not sure what is the goal here, something like session id?

ezr-ondrej: The goal is just to have a way to correlate all the requests from a single wizard.
Meaning to have a way in the backend to say what requests came from single wizard session (so yeah something like a session, but just a Wizard session).

adiabramovitch: @amirfefer Do you have a different solution than using the useEffect?

amirfefer: there are some solutions to bypass this issue, i.e create a [custom hook](https://dev.to/ag-grid/react-18-avoiding-use-effect-getting-called-twice-4i9e) which trigger useEffect only once.

src/Components/ProvisioningWizard/index.js
```diff
@@ -24,6 +27,11 @@ const ProvisioningWizard = ({ onClose, image, ...props }) => {
 
   const { isLoading, error: sourcesError, sources: availableSources } = useSourcesForImage(image, { refetch: stepIdReached <= 2 });
 
+  useEffectOnce(() => {
+    axios.defaults.headers.common['Provisioning-Wizard-Id'] = uuidv4();
+    return () => delete axios.defaults.headers.common['Provisioning-Wizard-Id'];
```
lzap, line 32: Just for the record, I suggest a generic term that don’t have wizard in the name. Feel fee to just use "X-Correlation-Id" or something like "X-Custom-Id" or whatever is generic. This can be then used by any request in order to "tag" logs, not just by the wizard.

amirfefer (approved):
Thanks @adiabramovitch ! 👍 
Tested and works well, mounted and unmounted
one nitpick for next time is to move the custom hook to `/common/hooks` folder for easier future reuse.

amirfefer (approved):

PR #246, opened 2023-04-14
feat: filter source in the select

ezr-ondrej:
This improves how we fetch usable sources.
Uses the filtered sources for the Sources select.
Drops notion of Shared with source in the select.
Insted preselects the source when only single is available to use.

Fixes HMS-1625

This goes above what we have implemented in #238 and uses the same approach for the sources select.
IMO much supperior approach as Users can select only sources that are usable to launch the given image.
IMO this is a great UX improvemet :)

On top of implemented changes, I've changed tests, so test itself is able to set the context values directly.

### DIFF
--- a/src/Components/Common/Hooks/sources.js
+++ b/src/Components/Common/Hooks/sources.js
@@ -1,0 +1,69 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { useQuery, useQueries } from 'react-query';
+
+import { SOURCES_QUERY_KEY, SOURCE_UPLOAD_INFO_KEY } from '../../../API/queryKeys';
+import { fetchSourcesList, fetchSourceUploadInfo } from '../../../API';
+import { AWS_PROVIDER, AZURE_PROVIDER } from '../../../constants';
+
+export const imageProps = PropTypes.shape({
+  name: PropTypes.string,
+  id: PropTypes.string,
+  provider: PropTypes.string,
+  architecture: PropTypes.string,
+  sourceIDs: PropTypes.arrayOf(PropTypes.string),
+  accountIDs: PropTypes.arrayOf(PropTypes.string),
+}).isRequired;
+
+export const useSourcesData = (provider, { refetch = false } = {}) => {
+  const {
+    error,
+    isLoading,
+    data: sources,
+  } = useQuery([SOURCES_QUERY_KEY, provider], () => fetchSourcesList(provider), {
+    enabled: !!provider,
+    refetchInterval: refetch && 10000,
+  });
+  return { error, isLoading, sources };
+};
+
+const useSourcesUploadInfos = (provider, { refetch }) => {
+  const { sources, isLoading: isLoadingSources, error: sourcesError } = useSourcesData(provider, { refetch });
+
+  const uploadInfos = useQueries(
+    sources?.map((source) => ({ queryKey: [SOURCE_UPLOAD_INFO_KEY, source.id], queryFn: () => fetchSourceUploadInfo(source.id), retry: 1 })) || []
+  );
+
+  const isLoading = isLoadingSources || uploadInfos.some((info) => info.isLoading);
+  const infos = [];
+  !isLoading &&
+    sources?.forEach((source, i) => {
+      uploadInfos[i].isSuccess && infos.push({ ...uploadInfos[i].data, ...source });
+    });
+
+  return { isLoading, error: sourcesError, infos };
+};
+
+const providerSourceFilter = (image) => {
+  switch (image.provider) {
+    case AWS_PROVIDER:
+      return (info) => image.sourceIDs?.includes(info.id) || image.accountIDs?.includes(info.aws.account_id);
+    case AZURE_PROVIDER:
+      return (info) => image.uploadOptions?.source_id === info.id || info.azure?.subscription_id === image.uploadOptions?.subscription_id;
+    default:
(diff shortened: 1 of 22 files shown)


src/Components/ProvisioningWizard/helpers.js
```diff
@@ -1,63 +1,12 @@
...
   name: PropTypes.string,
   id: PropTypes.string,
-  provider: PropTypes.string,
+  provider: PropTypes.oneOf([AWS_PROVIDER, AZURE_PROVIDER, GCP_PROVIDER]),
   architecture: PropTypes.string,
   sourceIDs: PropTypes.arrayOf(PropTypes.string),
   accountIDs: PropTypes.arrayOf(PropTypes.string),
 }).isRequired;
-
-export const sourcesData = (image, { refetch = false } = {}) => {
```
ezr-ondrej, line 17: moved to Hooks and renamed to `useXXX` so it can be properly checked by linters.

src/Components/Common/Hooks/sources.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+const useSourcesUploadInfos = (provider, { refetch }) => {
+  const { sources, isLoading: isLoadingSources, error: sourcesError } = useSourcesData(provider, { refetch });
+
+  const uploadInfos = useQueries(
+    sources?.map((source) => ({ queryKey: [SOURCE_UPLOAD_INFO_KEY, source.id], queryFn: () => fetchSourceUploadInfo(source.id), retry: 1 })) || []
+  );
+
+  const isLoading = isLoadingSources || uploadInfos.every((info) => info.isLoading);
```
amirfefer, line 37: should we use `some` instead of `every` ?  if just one  item from `uploadIndo` is loading, the general state is still loading, right?

```suggestion
  const isLoading = isLoadingSources || uploadInfos.some((info) => info.isLoading);
```

src/Components/Common/Hooks/sources.js
```diff
@@ -0,0 +1,70 @@
...
+  const infos = [];
+  sources?.forEach((source, i) => {
+    uploadInfos[i].isSuccess && infos.push({ ...uploadInfos[i].data, ...source });
+  });
```
amirfefer, line 41: no need to iterate if the queries are still loading, no?
```suggestion
  !isLoading && sources?.forEach((source, i) => {
    uploadInfos[i].isSuccess && infos.push({ ...uploadInfos[i].data, ...source });
  });
```

src/Components/Common/Hooks/sources.js
```diff
@@ -0,0 +1,70 @@
...
+  React.useEffect(() => {
+    if (!isLoading && !error) {
+      onSuccess(filteredSources);
+    }
+  }, [isLoading, error]);
+
+  if (isLoading || error) return { isLoading, error, sources: [] };
+  const filteredSources = image.isTesting ? infos : infos.filter(providerSourceFilter(image));
+
+  return { isLoading, error, sources: filteredSources };
```
amirfefer, line 69: nit: 
```suggestion
  const filteredSources = image.isTesting ? infos : infos?.filter(providerSourceFilter(image));
  return { isLoading, error, sources: filteredSources || [] };
```

src/Components/ProvisioningWizard/index.js
```diff
@@ -21,7 +22,7 @@ const ProvisioningWizard = ({ isOpen, onClose, image, ...props }) => {
   const [isConfirming, setConfirming] = React.useState(false);
   const [successfulLaunch, setLaunchSuccess] = React.useState();
 
-  const { isLoading, error: sourcesError, sources: availableSources } = sourcesForImage(image, { refetch: stepIdReached <= 2 });
+  const { isLoading, error: sourcesError, sources: availableSources } = useSourcesForImage(image, { refetch: stepIdReached <= 2 });
```
amirfefer, line 25: why do we refetch every 10 seconds?

src/Components/SourcesSelect/index.js
```diff
@@ -1,41 +1,33 @@
...
-  } = useQuery(sourcesQueryKey(provider), () => fetchSourcesList(provider), {
-    enabled: !!provider,
-    onSuccess: (data) => {
-      const id = chosenSource;
-
-      if (IB_SOURCE_PROVIDERS.includes(provider) && imageSourceID) {
-        setSelected(selectObject(imageSourceID, data.find((source) => source.id === imageSourceID).name));
+  const { isLoading, error, sources } = useSourcesForImage(image, {
+    onSuccess: (sources) => {
+      console.log(chosenSource, sources);
```
amirfefer, line 20: debugging leftovers?

amirfefer (commented):
Thanks @ezr-ondrej ! 👍  overall looks great :) 
I left a few comments

src/Components/ProvisioningWizard/index.js
ezr-ondrej: We want to refetch to cover the case where user had filled created a Source in a tab we open after clicking the "Create Source" button.

src/Components/Common/Hooks/sources.js
ezr-ondrej: sure :)

src/Components/SourcesSelect/index.js
ezr-ondrej: right xD thanks :)

amirfefer (approved):
Thanks @ezr-ondrej, tested and works well 👍

PR #249, opened 2023-04-19
Remove `isOpen` from Wizard

shaded-enmity:
PF4 wizard currently doesn't delegate the `appendTo` property so that it can be anchored to proper element. This removes the `isOpen` property so as not to force modal window. Callers of the Wizard need to wrap it in their own modal.

PF4 issue: https://github.com/patternfly/patternfly-react/issues/8972

### DIFF
--- a/src/Components/ProvisioningWizard/index.js
+++ b/src/Components/ProvisioningWizard/index.js
@@ -16,7 +16,7 @@ const DEFAULT_STEP_VALIDATION = {
   awsStep: false,
 };
 
-const ProvisioningWizard = ({ isOpen, onClose, image, ...props }) => {
+const ProvisioningWizard = ({ onClose, image, ...props }) => {
   const [stepIdReached, setStepIdReached] = React.useState(1);
   const [stepValidation, setStepValidation] = React.useState(DEFAULT_STEP_VALIDATION);
   const [isConfirming, setConfirming] = React.useState(false);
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
Both commit message and code linters have failed :(

amirfefer:
My concern is that the wizard component won't unmount on close (just be hidden), and will efect the wizard inner state.

shaded-enmity:
@amirfefer My patch in IB renders the wizard conditionally based on `isOpen` in the parent component, i.e.

`wizardOpen && <Wizard ...>` which will mount/unmount properly. Is there any other place where the wizard is invoked?

amirfefer:
IB is the only consumer ATM, if so it sounds safe, I'll test it with your patch. Thanks

ezr-ondrej:
I'm guessing we will need to fix the sample page for our IQE pipeline ....

ezr-ondrej (approved):
I guess the sample page works, I believe this is ready once the IB PR is in :+1:

shaded-enmity:
Can we merge this one first? I want to get the styling on the modal correct, but have no idea how to test without this being merged first :/ is there a simple way for me to wire up this branch with my branch in IB and test it together locally?

amirfefer:
This needs a quick rebase :)
I don't mind to merge this first, (@ezr-ondrej ? )  though without your patch, provisioning wizard might be broken in stage.

BTW, You can run  IB and provisiong wizard dev env together  with this script - https://github.com/RedHatInsights/image-builder-frontend/pull/873 :)

shaded-enmity:
@amirfefer thanks for the heads up, I've rebased the PR.

PR #250, opened 2023-04-26
fix: sources loading when one source failing

ezr-ondrej:
Mainly simplifies the step decision, so we dont flip rendered wizard steps there and back.

Adds tests for the form itself, for now only tests the loading and empty state. Fixes import of the icons.

Fixes HMS-1657

### DIFF
--- a/src/Components/ProvisioningWizard/steps/index.js
+++ b/src/Components/ProvisioningWizard/steps/index.js
@@ -60,6 +60,6 @@ const wizardSteps = ({ stepIdReached, image, stepValidation, setStepValidation, 
   },
 ];
 
-const steps = (props) => (!props.isLoading && props.availableSources.length > 0 ? wizardSteps(props) : missingSource(props));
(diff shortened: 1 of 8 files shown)


src/Components/ProvisioningWizard/steps/index.js
```diff
@@ -60,6 +60,6 @@ const wizardSteps = ({ stepIdReached, image, stepValidation, setStepValidation,
   },
 ];
 
-const steps = (props) => (!props.isLoading && props.availableSources.length > 0 ? wizardSteps(props) : missingSource(props));
+const steps = (props) => (props.availableSources.length > 0 ? wizardSteps(props) : missingSource(props));
```
ezr-ondrej, line 63: this is the main (and only) fix.

akhil-jha (approved):
I think I was able to reproduce the issue with bad sources. This fix worked! :tada: 
Let's see it over the stage now?

adiabramovitch:
let's give it a try on stage 
@ezr-ondrej Thank you

PR #254, opened 2023-04-27
ref(HMS-1573): fix failing tests

amirfefer:


ezr-ondrej:
nice, if we could fix the queryKeys, it would be great.
The commit message should be fix(...) or refs(...)

ezr-ondrej (approved):
Thanks @amirfefer, I guess this should fix it :+1:

PR #258, opened 2023-05-04
fix(HMS-1748): use staleTime config for preventing multiple requests

amirfefer:
Requests from client side are triggered multiple times due to missing configuration in react-query , data becomes stale as soon as it fetched, for further configuarion defaults of react-query - https://tanstack.com/query/v4/docs/react/guides/important-defaults 

This PR adds a global `staleTime` of 10 seconds, which is still aggresive policy,  but cancel it during resevation polling.

 I recored the amount of requests during  wizard steps progress before polling:

before:

https://user-images.githubusercontent.com/11807069/236284179-6f4f56ed-66f9-4908-87f3-ffbda47e7d5e.mov

after:

https://user-images.githubusercontent.com/11807069/236284230-7aa9f1cd-35bb-44af-9af4-3928140538dd.mov

More about `staleTime`, this [post](https://www.codemzy.com/blog/react-query-cachetime-staletime) covers it well

### DIFF
--- a/src/Components/Common/Query/index.js
+++ b/src/Components/Common/Query/index.js
@@ -3,7 +3,13 @@
 import React from 'react';
 import { QueryClient, QueryClientProvider } from 'react-query';
 
-const queryClient = new QueryClient();
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      staleTime: 5000,
+    },
+  },
+});
 
 const APIProvider = ({ children }) => <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>;
 
--- a/src/Components/InstanceTypesSelect/index.js
+++ b/src/Components/InstanceTypesSelect/index.js
@@ -20,6 +20,7 @@ const InstanceTypesSelect = ({ setValidation, architecture }) => {
     error,
     data: instanceTypes,
   } = useQuery(['instanceTypes', chosenRegion], () => fetchInstanceTypesList(chosenRegion, provider), {
+    staleTime: 5 * (60 * 1000), // data is considered fresh for 5 minutes (same as cacheTime)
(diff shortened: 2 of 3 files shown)


lzap:
Is this post-summit patch? If ain't broken...

src/Components/Common/Query/index.js
```diff
@@ -3,7 +3,13 @@
 import React from 'react';
 import { QueryClient, QueryClientProvider } from 'react-query';
 
-const queryClient = new QueryClient();
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      staleTime: 10000,
```
ezr-ondrej, line 9: I'd keep the default even a bit more agressive personally :)
We never know what data are we going to fetch in the future.
```suggestion
      staleTime: 5000,
```

ezr-ondrej (commented):
I'm not sure, but should we disable  this (or set explicit value) for Source fetching as well?
We rely on refetches (polling) when no source is available to figure out that User defined a source in a different tab.

I'd agree with Lukas here on the timeline, I agree we should do this and love the approach, but it's still cache invalidation, which might bring unexpected consequences, so I'd not merge before code freeze :)

amirfefer:
@ezr-ondrej, about the source polling, the refetch interval is 10 seconds - 
https://github.com/RHEnVision/provisioning-frontend/blob/6223e136432a09d0e399cd0100faa3eeb8013421/src/Components/Common/Hooks/sources.js#L15
as you can see, it does not override it.

I agree it is safer to merge post-summit 👍

ezr-ondrej:
So lets get that in.
@amirfefer what is the resolution for 10 vs 5 sec? I'm inclined to 5, but I'm happy to merge either just let me know :)

amirfefer:
lets start with 5 seconds globally :)
updated

PR #259, opened 2023-05-04
feat(HMS-1411):  ed25519 type is unsupported in azure

amirfefer:
This PR adds a validation error when ed25519 public key foramt is picked up from existing keys select or when uploading a new key.

### DIFF
--- a/src/Components/ProvisioningWizard/steps/Pubkeys/NewKeyForm.js
+++ b/src/Components/ProvisioningWizard/steps/Pubkeys/NewKeyForm.js
@@ -1,17 +1,22 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import { FormGroup, TextInput, FileUpload } from '@patternfly/react-core';
+import { FormGroup, TextInput, FileUpload, Button, Popover } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+
 import { useWizardContext } from '../../../Common/WizardContext';
+import { UNSUPPORTED_KEYS } from './helpers';
 
 // This is a simple regex format for public ssh key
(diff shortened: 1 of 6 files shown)


lzap:
LGTM I haven’t tested this.

src/Components/ProvisioningWizard/steps/Pubkeys/NewKeyForm.js
```diff
@@ -109,15 +112,13 @@ const NewSSHKeyForm = ({ setStepValidated }) => {
         />
       </FormGroup>
       <FormGroup
```
ezr-ondrej, line 114: Could we add the `?` when there are some unsuported keys?
When I select a `ed25519` key, I just get "is invalid or unsupported", but as a user, I might not know what does it mean.
If we would have the popup help when the provider is Azure (I guess this is ok to do statically) saying "Azure does not support ed25519 keys" and link to Azure statement about this, for example: https://learn.microsoft.com/en-us/troubleshoot/azure/virtual-machines/ed25519-ssh-keys I guess users would appreciate this (as it was very unexpected information for me, so I guess it will surprise many users and we don't want to get the blame if it's not our fault ;) )

ezr-ondrej (commented):
Love it! :orange_heart:
If we could squeeze in a helper popup as commeted below, it would be great, but if not, I'm good to merge AS-IS as this is still quite an improvement :+1:

amirfefer:
Added `?` helper :

Screen Shot 2023-05-07 at 15 30 57

ezr-ondrej (approved):
I'm happy, hopefully the pipeline just timeouted, if it goes green, lets merge.

Thanks @amirfefer ! This really helps with the UX for Azure :tada:

PR #260, opened 2023-05-05
build: clean unused files

amirfefer:
This cleans up unused files, and adds a test for the wizard sample page which consumed by integration tests


ezr-ondrej (approved):
Thanks, nice work, it's great to have less code to maintain and more tests :orange_heart:  :+1:

PR #261, opened 2023-05-08
fix: Sources link respect preview toggle

ezr-ondrej:
Currently we are only in preview, but we want to respect the toggle.

I'm keeping the change as simple as possible, in the future, we will want to do this dynamically.

Fixes HMS-1757


amirfefer (approved):
thanks @ezr-ondrej 👍

PR #262, opened 2023-05-17
Fix region select disabled in AWS provisioning wizard

tpapaioa:
If an AWS in image-builder has been cloned to one other region, then the region select dropdown is disabled in theprovisioning wizard, and the user can't select the 2nd region. This is only an issue when the image has been cloned only once; if the image is shared to a third region, then the dropdown is enabled.

This PR fixes that issue so that the dropdown is always enabled.

### DIFF
--- a/src/Components/RegionsSelect/RegionsSelect.test.js
+++ b/src/Components/RegionsSelect/RegionsSelect.test.js
@@ -26,19 +26,6 @@ describe('RegionSelect', () => {
     const filteredRegion = screen.queryByText(clonedImages.data[0].request.region);
     expect(filteredRegion).not.toBeInTheDocument();
   });
-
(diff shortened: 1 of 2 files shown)


ezr-ondrej:
We have a test case for the disable state: https://github.com/RHEnVision/provisioning-frontend/blob/6029c5ba688bb4ccf0132fb82a85ce42f26379bb/src/Components/RegionsSelect/RegionsSelect.test.js#L40

We will need to adjust that one maybe just by negating it? or deleting entirely

amirfefer (approved):
LGTM

Thanks @tpapaioa 👍 
@ezr-ondrej, selection is also enable for choosing another source, even if there is only one source, so this change makes sense to me.

PR #263, opened 2023-05-18
feat(HMS-1907):  add dev script for containerizing provisioning env

amirfefer:
is needed for https://github.com/RHEnVision/provisioning-backend/pull/539

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,10 @@ This project is based on react and uses the `console.redhat.com` micro-frontend 
 For further reading see [dev](https://github.com/RHEnVision/provisioning-frontend/blob/main/docs/dev.md)
 
 ## Setup
+
+### Containerized development environment
+
+For a containerized development environment, including provisioning-backend, db, redis, kafka and other integrated services, for further guidance, go to the [provisioning-compose](https://github.com/RHEnVision/provisioning-compose) repo.
 ### Initial etc/hosts setup
 
 In order to access the https://[env].foo.redhat.com in your browser, you have to add entries to your `/etc/hosts` file. This is a **one-time** setup that has to be done only once (unless you modify hosts) on each machine.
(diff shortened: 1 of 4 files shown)


ezr-ondrej:
Could we add it to a different directory?
Our pipelines currently build container from this code base (we are not using it yet, but it's getting built).
The pipeline allows for overriding the default build container and this would override it, I suspect that's the reason for pipeline failure.

Dockerfile.dev
```diff
@@ -0,0 +1,12 @@
+FROM docker.io/node:16
```
ezr-ondrej, line 1: Docker repositories are bit restrictive at times, would it be possible to use `registry.access.redhat.com/ubi8/nodejs-16`  for this?

I've tried to figure out what https://github.com/RedHatInsights/insights-frontend-builder-common/blob/be2fa6ca7dd7a84f8efce9cdd93febf8e541a5ce/src/frontend-build.sh#L145 does, but I failed to run it localy :/

lzap: I also think we should stick with Red Hat certified containers. All of our dependencies should work as they also use them.

lzap:
This is missing documetnation, if we want to take this route this must be fully documented ideally with a link from the main README.

lzap:
WFM TY

lzap:
This can be moved to https://github.com/lzap/provisioning-compose too

amirfefer:
@lzap, the Dockerfile was moved to [provisioning-compose](https://github.com/lzap/provisioning-compose), but the `dev` script is needed here in the `package.json`, it also can be used outside the container.

ezr-ondrej (approved):
Not at all I did not noticed we moved the compose parts already either, sorry! :(

This looks great, no comments :+1:

PR #268, opened 2023-06-11
fix(HMS-1917): Add source filtering for GCP

adiabramovitch:
Depends on https://github.com/RHEnVision/provisioning-backend/pull/571
- A function for filtering GCP sources was missing so I added it.
- a general link for GCP was added as well, it needs to be refactored to direct the customer to his specific project, I opened a bug for it. 
- Instances table will be added next
For now, we are relying on getting from the backend the email that the image was shared with.

### DIFF
--- a/src/Components/Common/Hooks/sources.js
+++ b/src/Components/Common/Hooks/sources.js
@@ -3,7 +3,7 @@ import { useQuery, useQueries } from 'react-query';
 
 import { SOURCES_QUERY_KEY, SOURCE_UPLOAD_INFO_KEY } from '../../../API/queryKeys';
 import { fetchSourcesList, fetchSourceUploadInfo } from '../../../API';
-import { AWS_PROVIDER, AZURE_PROVIDER } from '../../../constants';
+import { AWS_PROVIDER, AZURE_PROVIDER, GCP_PROVIDER } from '../../../constants';
 
 export const useSourcesData = (provider, { refetch = false } = {}) => {
   const {
@@ -40,6 +40,8 @@ const providerSourceFilter = (image) => {
       return (info) => image.sourceIDs?.includes(info.id) || image.accountIDs?.includes(info.aws.account_id);
     case AZURE_PROVIDER:
       return (info) => image.uploadOptions?.source_id === info.id || info.azure?.subscription_id === image.uploadOptions?.subscription_id;
+    case GCP_PROVIDER:
+      return () => true;
     default:
       return (info) => image.sourceIDs?.includes(info.id);
   }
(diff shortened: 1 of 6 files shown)


src/Components/ProvisioningWizard/steps/SourceMissing/DirectProviderLink.js
```diff
@@ -30,6 +30,10 @@ const DirectProviderLink = ({ image }) => {
         '/providers/Microsoft.Compute/images/' +
         uploadStatus.options.image_name;
       break;
+    case GCP_PROVIDER:
+      text = 'Launch with Google Cloud console';
+      url = 'https://console.cloud.google.com/welcome';
```
ezr-ondrej, line 35: I see image builder didn't find a way to launch the image directly either, but can we do better?
If I'm send to the welcome page as a user, I do not feel very comfortable as I might not know what to do next, maybe we can include similar info to what IB did include https://github.com/RedHatInsights/image-builder-frontend/blob/b73b0ecd564f02c5e738e76fab70c4a4d29606e9/src/Components/ImagesTable/ImageLinkDirect.js#L112

you know two lines "click link below and then search for the image by ...", but giving user at least some guidance :)

src/Components/InstancesTable/helpers.js
```diff
@@ -1,11 +1,13 @@
...
+import { AWS_PROVIDER, AZURE_PROVIDER, GCP_PROVIDER } from '../../constants';
 
 export const SSHUsername = (provider) => {
   switch (provider) {
     case AWS_PROVIDER:
       return 'ec2-user';
     case AZURE_PROVIDER:
       return 'azureuser';
+    case GCP_PROVIDER:
+      return 'gcpuser';
```
amirfefer, line 10: nit: this is needed for the follow-up PR, adding the instances table. but let's keep this

amirfefer (dismissed):
Thanks @adiabramovitch 👍 LGTM
This doesn't seem to  affect pr_checks because it is just GCP specific addition

ezr-ondrej:
I'm all for merging, let's skip the pipeline, if @adiabramovitch promises to take a look at failures we might find once the pipeline gets fixed.

ezr-ondrej (approved):

src/Components/Common/Hooks/sources.js
```diff
@@ -40,6 +40,8 @@ const providerSourceFilter = (image) => {
       return (info) => image.sourceIDs?.includes(info.id) || image.accountIDs?.includes(info.aws.account_id);
     case AZURE_PROVIDER:
       return (info) => image.uploadOptions?.source_id === info.id || info.azure?.subscription_id === image.uploadOptions?.subscription_id;
+    case GCP_PROVIDER:
+      return (info) => image.accountIDs?.includes(info.gcp.service_account_principal);
```
ezr-ondrej, line 44:
```suggestion
      return () => true;
```

ezr-ondrej (commented):
As discussed offline, now when we have all images shared with the GCP account, can we just not filter the GCP sources, as all can launch any image, right? :stuck_out_tongue:

ezr-ondrej (approved):
:tada: easy, I love simple PRs :tada:

PR #273, opened 2023-06-20
fix(HMS-1524): enable case insensitive search

ezr-ondrej:
Make the filter of InstanceType case insensitive


lzap:
Tests are red.

ezr-ondrej:
And will be until we fix'em :( the pipeline is still broken. Let me check that out.

lzap:
@akhil-jha do you think the pipeline is now ready

/retest

adiabramovitch:
@akhil-jha Is the pipeline fixed?

PR #275, opened 2023-06-25
fix(HMSPROV-2012): fix missing instances table

amirfefer:
In some cases, a polled reservation might be set at its final step (i.e. step 3 out of 3), but the `success` flag is still null, this caused the frontend app to end up polling.

polled reservation:
```
{
    "id": 25,
    "provider": 2,
    "created_at": "2023-06-25T18:33:20.686045Z",
    "steps": 3,
    "step_titles": [
        "Ensure public key",
        "Launch instance(s)",
        "Fetch instance(s) description"
    ],
    "step": 3,
    "status": "Instance(s) description fetched",
    "error": "",
    "finished_at": null,
    "success": null
}
```

### DIFF
--- a/src/Components/ProvisioningWizard/steps/ReservationProgress/index.js
+++ b/src/Components/ProvisioningWizard/steps/ReservationProgress/index.js
@@ -91,7 +91,7 @@ const ReservationProgress = ({ setLaunchSuccess }) => {
 
   // polling request
   const { data: polledReservation } = useQuery(['reservation', reservationID], () => fetchReservation(reservationID), {
-    enabled: !!reservationID && currentStep < steps.length && !currentError,
+    enabled: !!reservationID && !currentError,
     staleTime: 0, // disable cache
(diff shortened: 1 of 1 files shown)


src/Components/ProvisioningWizard/steps/ReservationProgress/index.js
```diff
@@ -91,7 +91,7 @@ const ReservationProgress = ({ setLaunchSuccess }) => {
 
   // polling request
   const { data: polledReservation } = useQuery(['reservation', reservationID], () => fetchReservation(reservationID), {
-    enabled: !!reservationID && currentStep < steps.length && !currentError,
+    enabled: !!reservationID && !currentError,
```
amirfefer, line 94: the `refetch` stops when the success/error flags are set, so no need to have this condition here. just caused this edge case

ezr-ondrej (approved):
Nice find! Thank you!

PR #278, opened 2023-07-03
feat(HMS-2022): Show DNS column only if public_dns(s) are not empty

adiabramovitch:
for GCP and Azure public DNS are not defined, this PR exposes the DNS column only if at least one launched instance has a public DNS defined. 
image

### DIFF
--- a/src/Components/InstancesTable/index.js
+++ b/src/Components/InstancesTable/index.js
@@ -30,6 +30,7 @@ const InstancesTable = ({ reservationID, provider, region }) => {
   };
 
   const instancesPerPage = instances?.slice(paginationOptions.startIdx, paginationOptions.endIdx);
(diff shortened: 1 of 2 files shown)


ezr-ondrej (commented):
Looks sane, I wonder if users won't get confused about missing column (thinking that it's broken thus missing), but I guess that should be understandable, especially given this is just quick info, long term we want Users to look for info in the inventory, right?
Did you think about that, what do you think?

adiabramovitch:
> Looks sane, I wonder if users won't get confused about missing column (thinking that it's broken thus missing), but I guess that should be understandable, especially given this is just quick info, long term we want Users to look for info in the inventory, right? Did you think about that, what do you think?

I think it is a temporary fix, seeing an empty column does not make much sense IMO, so I think this solution is good for now.

PR #279, opened 2023-07-03
feat(HMS-1930): Add image command clipboard for GCP

adiabramovitch:
This PR adds launch instance command clipboard to GCP Empty state when there aren't any GCP sources available

### DIFF
--- a/src/Components/GCPConsoleLaunch/GCPConsoleLaunch.scss
+++ b/src/Components/GCPConsoleLaunch/GCPConsoleLaunch.scss
@@ -1,0 +1,3 @@
+.custom-clipboard-copy {
+    width: 600px;
+  }
--- a/src/Components/GCPConsoleLaunch/index.js
+++ b/src/Components/GCPConsoleLaunch/index.js
@@ -1,0 +1,39 @@
+import React from 'react';
+import { ClipboardCopy, StackItem } from '@patternfly/react-core';
+import { imageProps } from '../ProvisioningWizard/helpers';
+import './GCPConsoleLaunch.scss';
+
+const launchInstanceCommand = (options) => {
+  return `gcloud compute instances create ${options.image_name}-instance --image-project ${options.project_id} --image ${options.image_name}`;
+};
+
+const GCPConsoleLaunch = ({ image }) => {
+  return (
+    <>
+      <br />
+      <p>or</p>
+      <br />
+      <StackItem>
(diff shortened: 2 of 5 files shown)


src/Components/ImageCommands/index.js
```diff
@@ -0,0 +1,31 @@
...
+import { ClipboardCopy, Stack, StackItem } from '@patternfly/react-core';
+import { imageProps } from '../ProvisioningWizard/helpers';
+
+const launchInstanceCommand = (options) => {
+  return `gcloud compute instances create ${options.image_name}-instance --image-project ${options.project_id} --image ${options.image_name}`;
+};
+
+const ImageCommand = ({ image }) => {
+  return (
+    <Stack>
```
amirfefer, line 11: This is already mounted in a `Stack` layout  in `SourceMissing` component, instead of having two stacks, instead you can wrap these `StackItems` with `<>` (react fragment)

src/Components/ImageCommands/index.js
```diff
@@ -0,0 +1,31 @@
...
+      <StackItem>
+        <br />
+        <strong> Launch with Google Cloud Console </strong>
+      </StackItem>
+      <StackItem isFilled>
+        <div style={{ width: '600px' }}>
```
amirfefer, line 17: can you please use css class instead?
the `ClipboardCopy` component also accepts classnames

src/Components/ImageCommands/index.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { ClipboardCopy, Stack, StackItem } from '@patternfly/react-core';
+import { imageProps } from '../ProvisioningWizard/helpers';
+
+const launchInstanceCommand = (options) => {
+  return `gcloud compute instances create ${options.image_name}-instance --image-project ${options.project_id} --image ${options.image_name}`;
+};
+
+const ImageCommand = ({ image }) => {
```
amirfefer, line 9: nit: The component's name is a bit ambiguous to me, maybe `GCPConsoleLaunch` or similar

amirfefer (changes requested):
Thanks @adiabramovitch, looks great, just a few comments

lzap:
Great idea, we should have it for all providers.

amirfefer (approved):
Thanks @adiabramovitch  👍

PR #283, opened 2023-07-16
fix(HMS-2114): Disable zone selection for GCP images

adiabramovitch:


src/Components/RegionsSelect/index.js
```diff
@@ -71,6 +71,7 @@ const RegionsSelect = ({ provider, currentRegion, composeID, onChange }) => {
       selections={currentRegion}
       onToggle={onToggle}
       onSelect={onSelect}
+      isDisabled={provider === GCP_PROVIDER}
```
amirfefer, line 74: AFAIK also in Azure users can't clone to another region, no?

ezr-ondrej: correct, we are now not disabling it for Azure as well, so it's broken there as well.

```suggestion
      isDisabled={provider !== AWS_PROVIDER}
```

Or ideally have a new helper function

```suggestion
      isDisabled={!supportRegionCustomization(provider)}
```

adiabramovitch: TY, I was not sure, added a fix.

ezr-ondrej (approved):
Great! :orange_heart:

PR #284, opened 2023-07-18
refactor: Disable templates dropdown when templates array is empty

adiabramovitch:


amirfefer (approved):
Nice!
Thanks @adiabramovitch 👍

PR #285, opened 2023-07-18
feat(HMS-2176): add info banner for notification

amirfefer:
Screen Shot 2023-07-18 at 18 16 25

### DIFF
--- a/src/Components/Common/Hooks/useLocalStorage.js
+++ b/src/Components/Common/Hooks/useLocalStorage.js
@@ -1,0 +1,29 @@
+import { useState } from 'react';
+
+function useLocalStorage(key, initialValue) {
+  const storedValue = () => {
+    try {
+      const item = window.localStorage.getItem(key);
+      return item ? JSON.parse(item) : initialValue;
+    } catch (error) {
+      console.log(error);
+      return initialValue;
+    }
+  };
+
+  const [value, setValue] = useState(storedValue);
+
+  const setStoredValue = (value) => {
+    try {
+      const valueToStore = value instanceof Function ? value(storedValue) : value;
+      setValue(valueToStore);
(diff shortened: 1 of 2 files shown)


lzap:
Looks a bit too heavy for every single launch to me.

Idea: We could show this only in case a launch exceeds our wait time.

amirfefer:
Maybe we can add a dismiss button?
But I think it more like a UX question, @MariSvirik, opinion?

ezr-ondrej:
> for preventing mixing with production once it merged, would it be better to have a feature flag?

I'd say so :) ideally we would expose the backend flag, but that's overkill.
another solution would be to merge this once we are flipping the backend flag on, it's not great, but it is a solution :)

MariSvirik:
dismiss button would work. But if the notifications are already set, we don't show it, right?
Which flag are we talking about?

ezr-ondrej:
> Which flag are we talking about?

on backend we have a flag controling the notification integration being on or off (and it's off on prod now), so we need to make sure this notice is not enabled before we enable the notifications in production :)

adiabramovitch:
Do we have a decision here?

ezr-ondrej:
I think we decided to wait for the notifications to be enabled on prod and then merge this banner.

BTW @amirfefer while working on updating the banner in IB, I've noticed they are enabling users to close the banner, I guess that's not an issue for now, but might be something to think of as a next step if we want to keep the banner longer, it would be nice to give Users way to opt out from seeing it :)

ezr-ondrej (dismissed):
I'm fine merging this for now :)
Or @amirfefer did you want to work on the closing?

amirfefer:
Added a dismiss button.

ezr-ondrej (approved):
Works as expected! Thank you @amirfefer ! :tada:

PR #291, opened 2023-08-01
fix(HMS-1757): remove hard coded preview in sources link

amirfefer:


adiabramovitch (approved):
LGTM tested it and nothing seems to change
Thanks!

PR #293, opened 2023-08-03
fix(HMS-2256): enhance wizard first page interface

amirfefer:
This PR enhances the Account and customization step in the wizard

- All select inputs have `maxHeight` for fixing select content override
- Content is aligned to the wizard layout, minimizes the need for a scroll bar
- Alert warning for the instance types is removed and replaced by inner validation and helper text
- Instance type select switched location with Templates select
- Template select's dropdown opens in the upper direction

Before:

https://github.com/RHEnVision/provisioning-frontend/assets/11807069/1a5e027e-286b-4d31-83f0-518b805b719c

After:

https://github.com/RHEnVision/provisioning-frontend/assets/11807069/034ff612-668d-4744-b395-29089511c092

### DIFF
--- a/src/Components/InstanceTypesSelect/InstanceTypesSelect.test.js
+++ b/src/Components/InstanceTypesSelect/InstanceTypesSelect.test.js
@@ -40,8 +40,8 @@ describe('InstanceTypesSelect', () => {
       await mountSelectAndClick();
       const unsupportedInstance = await screen.findByText('t1.nano');
       await userEvent.click(unsupportedInstance);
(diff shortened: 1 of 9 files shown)


lzap:
Please note this is a temporary solution. Long term, we would like to break this page into multiple screens or something like that.

WFM @ezr-ondrej ?

Can we merge today so we can hit prod tomorrow?

ezr-ondrej (approved):
Looks great to me :orange_heart: 
I do not fully understand the options, but looks quite sane :+1:

I'm happy to merge if this is not breaking the pipeline :+1:

PR #295, opened 2023-08-03
ci: Update image UUIDs

akhil-jha:
Older images have been removed


mshriver (approved):

mshriver:
preview isn't enabled when the CJI is running so the launch button isn't as expected. Unrelated to the image change, and the test pipeline screenshots show the image loaded into the DB.

PR #297, opened 2023-08-07
ci: use older build container for pr_check

ezr-ondrej:
The build container is now building into dist/stable folder. For PR_CHECK it does not tho, but the Caddyfile is still referencing this new folder.


ezr-ondrej:
This won't work, the BUILD_TAG is not parametrizable

PR #300, opened 2023-08-08
fix(HMS-2297): clear key format error

amirfefer:
When an invalid format for Azure key is selected, and then a valid (RSA) key is selected, the validation is not cleared, so it still seems like the key is not valid.


ezr-ondrej (approved):
Thanks! :orange_heart:

PR #306, opened 2023-08-16
feat(HMS-2369): remove stage old deployment

amirfefer:
Part of the containerized frontend process.


ezr-ondrej (commented):
I guess we could also disable the automatical branch sync, as it would be syncing for no reason:

https://github.com/RHEnVision/provisioning-frontend/blob/def98124bc48e60e49849daba9b3825c6ff5bb06/.github/workflows/sync_branches.yml#L4

ezr-ondrej (approved):
Nice, thank you! :+1:

PR #307, opened 2023-08-17
feat(HMS-2378): refactor the template select alert

amirfefer:
This PR continues the refactoring for removing alerts into inline validations:

Screen Shot 2023-08-17 at 18 03 53


ezr-ondrej (approved):
Thanks @amirfefer ! :pray:

PR #308, opened 2023-08-22
ci: release to Prod stable on prod push

ezr-ondrej:
Push to prod stable as well as Prod beta


ezr-ondrej:
Failure not related, merging so we are ready :+1:

PR #313, opened 2023-08-27
fix(HMS-1913): add an optinal string to template select

amirfefer:
This PR Adds the word "Optional" to select templates field to better indicate that it is optional.

Screen Shot 2023-08-27 at 16 37 53

Another solution is to use patternfly Form's `additionalLabel` prop:

Screen Shot 2023-08-27 at 16 25 56

/cc @MariSvirik


lzap (approved):
I can review that! :-) LGTM

ezr-ondrej:
Yeah, let's merge, better then nothing, it's easy to change if Maria would like the second option more :))

PR #321, opened 2023-08-30
build: fix prod deploy

amirfefer:


ezr-ondrej (approved):
Heh, I did not check properly 🤦
Thanks for the quickfix! 🧡

PR #324, opened 2023-09-06
feat(HMS-1897): add export to csv btn to instances table

amirfefer:
This adds an export to csv button to the instances table

Screen Shot 2023-09-06 at 11 30 38

vs

Screen Shot 2023-09-05 at 18 51 09

### DIFF
--- a/src/Components/InstancesTable/helpers.js
+++ b/src/Components/InstancesTable/helpers.js
@@ -12,3 +12,29 @@ export const SSHUsername = (provider) => {
       '';
   }
 };
+
+const flattenInstances = (instances) => {
+  return instances.map(({ instance_id, detail }) => ({ id: instance_id, ...detail }));
+};
+const convertToCSV = (objArray) => {
+  const flatten = flattenInstances(objArray);
+  const header = Object.keys(flatten[0]);
+  const csv = flatten.map((row) => header.map((fieldName) => JSON.stringify(row[fieldName])).join(','));
+  csv.unshift(header.join(','));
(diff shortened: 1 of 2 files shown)


amirfefer:
I took the design from Patternfly guidelines: 
https://www.patternfly.org/2023.02/components/table/design-guidelines
The button and pagination are under the Toolbar component

We can use a kebab dropdown instead of a button action, but I don't think we have another action to add.

adiabramovitch:
I am not expert, but I think that something more minimal like a download button would suffice

MariSvirik:
Great job trying to find the answers in PF guidelines, but this was a tricky one.
I talked to the design team about this as I wasn't sure if my solution was correct. 
We ended up voting for this:
The button is located under the table - it follows an empty state pattern (actions are at the bottom) and also an eye pattern (content leading all the way down).  If you want we can talk about the button being secondary, but I don't think we want to make it that prominent. 
Also, be aware that an icon is different.
Screenshot 2023-09-06 at 15 19 56

amirfefer:
Thanks @MariSvirik,

My concern is when there are multiple instances to scroll in order to see the button at the bottom, but as you mention, there's a pagination for ease.

updated:
Screen Shot 2023-09-07 at 16 01 14

adiabramovitch (approved):
LGTM, checked and work as expected. waiting for the pipeline to pass.

ezr-ondrej:
Thank you @amirfefer ! :tada:

PR #325, opened 2023-09-06
build: use npm build instead of yarn

amirfefer:
in GitHub action, we use yarn for building, but we use npm.


ezr-ondrej (commented):
Makes sense :+1:

Tho I can't find the artifact, so I'd rather wait with this for the transition to the containerized deployments, lets not change the old way before we stop using it :)

ezr-ondrej (approved):

ezr-ondrej:
Thanks @amirfefer ! :)

PR #327, opened 2023-09-08
ci: Drop Sonarqube for Synk

ezr-ondrej:
Synk is much easier to work with and integrate with. We won't have coverage reports and other quality checks, but there are better tools for the job.


lzap (approved):
Tests failure, LGTM.

PR #333, opened 2023-09-10
feat(HMS-2542): remove travis deployment

amirfefer:
This removes the deployment configuration via travis and the frontend build repo.
Merge only after akamai transition is done for prod.


ezr-ondrej (approved):
LGTM, but as stated, needs to wait for Akamai! :)

ezr-ondrej:
Akamai switch was done, I've verified everything works.
Unfortunatelly I'm not able to verify that we  are really consuming assets from new deploy, but I guess we need to trust the system there ;P

We can always revert this tho :)

PR #334, opened 2023-09-12
feat(HMS-2227): Add Pubkeys pagination

adiabramovitch:
[Screencast from 2023-09-20 14-33-03.webm](https://github.com/RHEnVision/provisioning-frontend/assets/57755873/5fef7d96-82f3-44d7-8150-49a4c03032b7)

### DIFF
--- a/src/API/helpers.js
+++ b/src/API/helpers.js
@@ -1,2 +1,4 @@
 export const provisioningUrl = (path) => `/api/provisioning/v1/${path}`;
 export const imageBuilderURL = (path) => `/api/image-builder/v1/${path}`;
+export const LIMIT = 100;
+export const OFFSET = 0;
--- a/src/Components/ProvisioningWizard/steps/Pubkeys/PubkeySelect.js
+++ b/src/Components/ProvisioningWizard/steps/Pubkeys/PubkeySelect.js
@@ -1,13 +1,14 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Alert, Select, SelectOption, Spinner, FormGroup } from '@patternfly/react-core';
-import { useQuery } from '@tanstack/react-query';
+import { useInfiniteQuery } from '@tanstack/react-query';
 
 import { PUBKEYS_QUERY_KEY } from '../../../../API/queryKeys';
 import { fetchPubkeysList } from '../../../../API';
 import { useWizardContext } from '../../../Common/WizardContext';
 import { isNotSupportKeyFormat } from './helpers';
 import { humanizeProvider } from '../../../Common/helpers';
+import { LIMIT } from '../../../../API/helpers';
 
 const selectOptionObj = (id, name) => ({
   id: id,
@@ -34,7 +35,17 @@ const PubkeySelect = ({ setStepValidated }) => {
     }
   }, [selection]);
 
-  const { isLoading, isError, data: pubkeys } = useQuery(PUBKEYS_QUERY_KEY, fetchPubkeysList);
+  const { data, fetchNextPage, isLoading, isError, hasNextPage } = useInfiniteQuery({
+    queryKey: [PUBKEYS_QUERY_KEY],
+    queryFn: ({ pageParam = 0 }) => fetchPubkeysList(pageParam),
+    getNextPageParam: (lastPage) => {
+      const nextOffset = lastPage.data?.length < lastPage.metadata?.total ? lastPage.offset + LIMIT : false;
+      return nextOffset;
+    },
+  });
+
+  const total = data?.pages[0]?.metadata?.total || 0;
+  const pubkeys = data?.pages.map((page) => page.data).flat() || [];
 
   const onSelect = (event, value) => {
     setWizardContext((prevState) => ({
@@ -61,7 +72,7 @@ const PubkeySelect = ({ setStepValidated }) => {
 
   return (
     <FormGroup
-      helperTextInvalid={`Key format is not support in ${humanizeProvider(wizardContext.provider)}`}
+      helperTextInvalid={`Key format is not supported in ${humanizeProvider(wizardContext.provider)}`}
       label="Select public key"
       validated={!isKeySupported && 'error'}
     >
@@ -73,11 +84,25 @@ const PubkeySelect = ({ setStepValidated }) => {
         selections={selection}
         placeholderText="Select public key..."
         aria-label="Select public key"
-        validated={!isKeySupported && 'error'}
+        {...(pubkeys &&
+          hasNextPage &&
+          pubkeys.length < total && {
+            loadingVariant: {
+              text: `View more (${total - pubkeys.length})`,
+              onClick: () => {
+                fetchNextPage();
+              },
+            },
+          })}
       >
-        {pubkeys.map(({ id, name }) => (
-          <SelectOption aria-label={`Public key ${name}`} key={id} value={selectOptionObj(id, name)} />
-        ))}
+        {pubkeys &&
+          data.pages.map((group, i) => (
+            <React.Fragment key={i}>
+              {group.data.map((pubkey) => (
+                <SelectOption aria-label={`Public key ${pubkey.name}`} key={pubkey.id} value={selectOptionObj(pubkey.id, pubkey.name)} />
+              ))}
+            </React.Fragment>
+          ))}
       </Select>
     </FormGroup>
   );
--- a/src/Components/ProvisioningWizard/steps/Pubkeys/Pubkeys.test.js
+++ b/src/Components/ProvisioningWizard/steps/Pubkeys/Pubkeys.test.js
@@ -2,7 +2,7 @@ import React from 'react';
 import userEvent from '@testing-library/user-event';
 import Pubkeys from '.';
 
-import { render, screen } from '../../../../mocks/utils';
+import { render, screen, waitFor } from '../../../../mocks/utils';
 import { provisioningUrl } from '../../../../API/helpers';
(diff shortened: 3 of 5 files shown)


src/Components/SourcesSelect/index.js
```diff
@@ -1,35 +1,66 @@
...
+              chosenSource: sources[0].id,
+            }));
+            setValidation('success');
+          }
+          setSrcs((prevSrcs) => [...prevSrcs, ...sources]);
+        },
+      },
+      LIMIT,
+      offset
+    );
```
adiabramovitch, line 46: This block needs to e called every time the offset is being changed, meaning,every time the View more button was pressed. 
But as you can see isLoading, error are dependent at that call as well. 
So I thought to have a function that is dealing with the entire fetching logic and call it every time the offset was updated. 
The browser crushes and I guess it is not the correct way, for Pubkeys we do no have that problem.

adiabramovitch: @ezr-ondrej

src/Components/ProvisioningWizard/steps/Pubkeys/PubkeySelect.js
```diff
@@ -34,7 +37,13 @@ const PubkeySelect = ({ setStepValidated }) => {
     }
   }, [selection]);
 
-  const { isLoading, isError, data: pubkeys } = useQuery(PUBKEYS_QUERY_KEY, fetchPubkeysList);
+  const { isLoading, isError, data } = useQuery([PUBKEYS_QUERY_KEY, LIMIT, offset], () => fetchPubkeysList(LIMIT, offset));
```
ezr-ondrej, line 40: given the LIMIT is static, we might omit it from the key IMO :)

src/Components/ProvisioningWizard/steps/Pubkeys/Pubkeys.test.js
```diff
@@ -19,6 +19,33 @@ describe('Pubkeys', () => {
       expect(items).toHaveLength(2);
     });
 
+    test('view more with click', async () => {
```
ezr-ondrej, line 22: Wow, this seems to work, because the tests are green but I wonder ... HOW? xD
We have the limit set to 100 constant and we mock only the first request...
Do you know why it actually works? :open_mouth:

src/Components/ProvisioningWizard/steps/Pubkeys/PubkeySelect.js
```diff
@@ -73,11 +91,17 @@ const PubkeySelect = ({ setStepValidated }) => {
         selections={selection}
         placeholderText="Select public key..."
         aria-label="Select public key"
-        validated={!isKeySupported && 'error'}
+        {...(pubkeys &&
+          pubkeys.length < total && {
```
ezr-ondrej, line 95: could we use `hasNextPage` from the [useInfiniteQuery return values](https://tanstack.com/query/v4/docs/react/reference/useInfiniteQuery) for this condition? :thinking:

src/Components/ProvisioningWizard/steps/Pubkeys/Pubkeys.test.js
```diff
@@ -19,6 +19,33 @@ describe('Pubkeys', () => {
...
+      const { server, rest } = window.msw;
+      server.use(
+        rest.get(provisioningUrl('pubkeys?limit=1&offset=0'), (req, res, ctx) => {
```
adiabramovitch, line 25: Because Of that :)

src/Components/ProvisioningWizard/steps/Pubkeys/PubkeySelect.js
adiabramovitch: I tried but it didn't seem to work, but I'll try again maybe I missed something

src/Components/ProvisioningWizard/steps/Pubkeys/Pubkeys.test.js
ezr-ondrej: tho I'd expect the second request to be `limit=100&offset=100`. Tho maybe the mock ignores params! :) that might be it...
Then it's bit confusing to have the params in the URL tho :)
I still don't get the LIMIT working as expected tho (to be 1, when we have constant set to 100)

ezr-ondrej:
btw there is a typo in paginatin => pagination, but I can fix that on merge

akhil-jha:
Do you think it would be too much clicking on View More to get to your pubkey? What if the pubkey name starts with 'z'? 
Can we do it in a batch of 10?

src/Components/ProvisioningWizard/steps/Pubkeys/Pubkeys.test.js
akhil-jha: The test is green because, for each launch click, a new pubkey is created, and in the test teardown, the pubkey is removed.....so in a way, dropdown of pubkey has only 1 pubkey present. 
The test will fail when it comes to integration test since there might be pubkeys present of other users.

ezr-ondrej:
> Do you think it would be too much clicking on View More to get to your pubkey? What if the pubkey name starts with 'z'? Can we do it in a batch of 10?

I believe we are doing it in batch of 100 :) The example is just to  showcase it with lower value ;)

src/API/helpers.js
```diff
@@ -1,2 +1,3 @@
 export const provisioningUrl = (path) => `/api/provisioning/v1/${path}`;
 export const imageBuilderURL = (path) => `/api/image-builder/v1/${path}`;
+export const LIMIT = 100;
```
ezr-ondrej, line 3: @akhil-jha here we've set the final batch, tho agreed the example is missleading, but it would be hard to create 100+ keys to showcase it.

akhil-jha: Ahh. Okay. :)

adiabramovitch:
@ezr-ondrej I am not sure, is there a problem? :)

ezr-ondrej:
> @ezr-ondrej I am not sure, is there a problem? :)

I've stared an offline discussion, but I believe the test is still not clear, I'd like it to better explain the code behaviour.
[This discussion](https://github.com/RHEnVision/provisioning-frontend/pull/334#discussion_r1331583994) seems unresolved to me.

amirfefer (changes requested):
Thanks @adiabramovitch! tested and it works as expected.

On every initial rendering,  it looks like there's a redundant API call, with a given object as offset:
Screen Shot 2023-10-04 at 21 53 57

Also, no spinner or any indication for loading is shown during view more, like the patterfnly example, which I don't mind adding in a follow-up PR.

https://github.com/RHEnVision/provisioning-frontend/assets/11807069/907409c5-e634-4896-9607-8fd5bcc4d47f

src/Components/ProvisioningWizard/steps/Pubkeys/PubkeySelect.js
```diff
@@ -34,7 +35,21 @@ const PubkeySelect = ({ setStepValidated }) => {
...
+    getNextPageParam: (lastPage) => {
+      const nextOffset = lastPage.data?.length < lastPage.metadata?.total ? lastPage.offset + LIMIT : false;
+      return nextOffset;
+    },
+  });
+
+  const total = data?.pages.map((item) => item.metadata?.total || 0)[0];
+  const pubkeys = data?.pages.reduce((acc, page) => {
+    return [...acc, ...page.data];
+  }, []);
```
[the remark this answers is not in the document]
adiabramovitch: Fixed

adiabramovitch:
> Thanks @adiabramovitch! tested and it works as expected.
> 
> On every initial rendering, it looks like there's a redundant API call, with a given object as offset: Screen Shot 2023-10-04 at 21 53 57
> 
> Also, no spinner or any indication for loading is shown during view more, like the patterfnly example, which I don't mind adding in a follow-up PR.
> 
>  Screen.Recording.2023-10-04.at.21.58.32.mov

There is an extra call from here: frontend/src/Components/ProvisioningWizard/steps/Pubkeys/index.js 
It is solved when adding the offset parameter, it does not seem like the default value works.

src/Components/ProvisioningWizard/steps/Pubkeys/index.js
```diff
@@ -1,19 +1,25 @@
...
 import { fetchPubkeysList } from '../../../../API';
+import { OFFSET } from '../../../../API/helpers';
 
 const EXIST_KEY_OPTION = 'existKey';
 const NEW_KEY_OPTION = 'newKey';
 
 const PublicKeys = ({ setStepValidated }) => {
   const [wizardContext, setWizardContext] = useWizardContext();
-  const { isError, data: pubkeys } = useQuery(PUBKEYS_QUERY_KEY, fetchPubkeysList);
+  const { isError, data } = useInfiniteQuery({
```
ezr-ondrej, line 17: also when testing, I've discovered, that using `useQuery` and `useInfiniteQuery` under the same key trips the query weirdly, so I wanted to be on the safe side and migrated to usage of infiniteQuery here as well.

ezr-ondrej:
/retest

Oh my the bot misbehaves :(

ezr-ondrej:
/retest

:roll_eyes:

akhil-jha:
@ezr-ondrej  Finally you got the green xD

ezr-ondrej:
> @ezr-ondrej Finally you got the green xD

And the hell of a ride it was xD The bot just did not like testing today xD

ezr-ondrej (approved):
Thanks @adiabramovitch ! :orange_heart:

PR #336, opened 2023-09-14
build: Update dependencies

ezr-ondrej:


lzap:
Linter issues and conflicting file.

ezr-ondrej:
Yeah I don't think it will work, it's  just too much of a big bang and it's impossible to see what's being upgraded :face_exhaling:
So it actually breaks tests, but it's impossible (for me) to see why, so closing and will update selectively :shrug:

PR #341, opened 2023-09-20
fix(HMS-2618): format regions error correctly

ezr-ondrej:
Moves the error state on RegionSelect to use inline FormHelperText and error validations.

The fix itself is wrapping the RegionSelect in FormGroup.

BEFORE:
Screenshot from 2023-09-20 09-14-16

AFTER:
Screenshot from 2023-09-20 09-52-10


adiabramovitch (approved):
LGTM

ezr-ondrej:
Thanks! :orange_heart: definitelly better then what we have, I'm discussing with Maria how to make it even better :)

PR #345, opened 2023-10-04
Replace 'vCPU' with 'vcpus'

stejskalleos:
GCE shows `vCPU` in the description of the machine, not the `vcpus`.


avitova:
Nice, thanks!
Could you please edit the commit message based on our [docs](https://github.com/RHEnVision/provisioning-backend/blob/main/CONTRIBUTING.md)? Sorry, it seems we do not have this document for front end, but the same rules should apply.

src/Utils/querySearch.js
```diff
@@ -5,6 +5,7 @@ const OPERATORS = ['<', '>', '='];
 const parseQuery = (query) => {
   query = query.replace('memory', 'memory_mib');
   query = query.replace('storage', 'storage_gb');
+  query = query.replace('vCPU', 'vcpus');
```
amirfefer, line 8: I'd suggest applying it also to `vcpu`, maybe use a case-insensitive query?

ezr-ondrej:
> Could you please edit the commit message based on our [docs](https://github.com/RHEnVision/provisioning-backend/blob/main/CONTRIBUTING.md)? Sorry, it seems we do not have this document for front end, but the same rules should apply.

@stejskalleos feel free to use `fix(HMS-2301): Replace 'vCPU' with 'vcpus'` it will satisfy the linter, even tho it's bit of a cheating to use closed ticket, for such a small change I'd be perfectly fine with it :)

ezr-ondrej:
I've addressed @avitova and @amirfefer concerns, so we can get this in, could you please re-review? :pray:

ezr-ondrej:
/retest

weird timeout :thinking:

ezr-ondrej:
I'll go ahead and merge, Amir might not get to it and we addressed his comments, so we should be good :)
I'm happy to address any additional comments, if Amir has some :)

PR #349, opened 2023-10-11
ci: test env

ezr-ondrej:


ezr-ondrej:
Yeah, the environment variables are just not there :/

PR #350, opened 2023-10-12
build: add basic tsconfig.json

ezr-ondrej:
The frontend config now requires a tsconfig for some reason. Probably update that got in too early.
This is just really basic tscript.json from template, to appease the plugin.

No idea why it's now needed, but shouldn't hurt :shrug:


ezr-ondrej:
Well, it breaks builds :thinking: xD

PR #351, opened 2023-10-17
InstanceTypesSelect: improve arch compatibility

lavocatt:
Previously:
The InstanceTypesSelect would only show results for an architecture if the architecture returned by the backend was exactly matching the one asked by the user.

This proved to be problematic because some architecures can have many names. For instance `aarch64` can be interchangeably used with `arm64`.

To to make the ProvisionningWizard more flexible, the instance selection dropdown now is using the same name mapping for the architectures as the one found in the backend. See:
https://github.com/RHEnVision/provisioning-backend/blob/0f3abc7328ebda3314d7496544b019a36a061152/internal/clients/architecture.go#L25

This work is related to HMS-1135

### DIFF
--- a/src/Components/InstanceTypesSelect/InstanceTypesSelect.test.js
+++ b/src/Components/InstanceTypesSelect/InstanceTypesSelect.test.js
@@ -4,7 +4,7 @@ import InstanceTypesSelect from '.';
 import { awsInstanceTypeList, azureInstanceTypeList } from '../../mocks/fixtures/instanceTypes.fixtures';
 import { render, screen } from '../../mocks/utils';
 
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
Nice! :)
IMO we are going bit overboard with allowing that many formats, but at least we are ready! :sunglasses:

lavocatt:
> Nice! :) IMO we are going bit overboard with allowing that many formats, but at least we are ready! 😎

Ahah, indeed, I thought you know, why not just get all of the mappings from the backend at once :smile:

ezr-ondrej:
Thanks! :orange_heart:

PR #353, opened 2023-10-24
feat(HMS-2255): deploy to image RG by default

ezr-ondrej:
Use image RG by default, we disable selection of resource group entirely  for now.

This also removes the disabled region selection, as we default to the  Resource Group location(region) and that's the best practice in Azure,  so we don't want to promote overriding it.

Previously:
Screenshot from 2023-10-24 14-48-38

Now:
Screenshot from 2023-10-24 14-47-56


ezr-ondrej:
Closing this one as per discussion around enabling customers to override resource group and it seems such feature is useful, so going with https://github.com/RHEnVision/provisioning-frontend/pull/354

PR #354, opened 2023-10-25
feat(HMS-2255): deploy to image RG by default

ezr-ondrej:
Use image RG by default and display clearly what the default is.

This also removes the region selection entirely, as we default to the Resource Group location(region) and that's the best practice in Azure, so we don't want to promote overriding it.

We have went to allowing Customers to shoot themselves in the foot a bit by selecting a wrong resource group, but we decided that's a better option, than disabling it completely, as we found customer value in customizing the RG.

Second version to #353 with enabling override for the resource group.

[Screencast from 2023-10-25 13-17-04.webm](https://github.com/RHEnVision/provisioning-frontend/assets/2884324/c4644daa-5d7c-46e4-9d69-9d1fdb1268ba)

### DIFF
--- a/src/Components/AzureResourceGroup/AzureResourceGroup.test.js
+++ b/src/Components/AzureResourceGroup/AzureResourceGroup.test.js
@@ -6,6 +6,11 @@ import { azureSourceUploadInfo } from '../../mocks/fixtures/sources.fixtures';
 import AzureResourceGroup from '.';
 
 describe('AzureResourceGroup', () => {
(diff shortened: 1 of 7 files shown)


akhil-jha (approved):

ezr-ondrej:
@amirfefer can you take a look pls? :)

ezr-ondrej:
Amir won't get to it anytime soon, I'm going to merge this.

PR #357, opened 2023-11-03
fix: do not show region in review for Azure provider

ezr-ondrej:
Removes region info from Review for Azure.
For Azure we use Resource Group insted.

image

Refs HMS-2255


ezr-ondrej:
the tests do not like this, are we doing some UI assertions?
```
assert response.location == location
AssertionError: assert '' == 'eastus_1'
```

ezr-ondrej:
/retest

Still public IP issue, I've filed IQE MR suggesting to wait for private IP instead.

akhil-jha:
Let's merge this PR. The failure is related to aws from testing end. That will be fixed in the plugin only. 
Need this PR for integration test now

akhil-jha (approved):

ezr-ondrej:
I concur! :)

PR #370, opened 2024-01-09
fix(HMS-3373):  update msw and frontend-components-config

amirfefer:
WIP

This PR update `msw` mocking server worker, which is used in our tests.
The msw (1.x -> 2.x) update is required for updating  `@redhat-cloud-services/frontend-components-config` due to typescript peer dependent.

### DIFF
--- a/jest.polyfills.js
+++ b/jest.polyfills.js
@@ -1,0 +1,31 @@
+// jest.polyfills.js
+/**
+ * @note The block below contains polyfills for Node.js globals
+ * required for Jest to function when running JSDOM tests.
+ * These HAVE to be require's and HAVE to be in this exact
+ * order, since "undici" depends on the "TextEncoder" global API.
+ *
+ * Consider migrating to a more modern test runner if
+ * you don't want to deal with this.
+ */
+
+const { TextDecoder, TextEncoder, ReadableStream } = require("node:util")
+
+Object.defineProperties(globalThis, {
+  TextDecoder: { value: TextDecoder },
+  TextEncoder: { value: TextEncoder },
+  ReadableStream: { value: ReadableStream },
+})
+
+const { Blob, File } = require("node:buffer")
+const { fetch, Headers, FormData, Request, Response } = require("undici")
+
+Object.defineProperties(globalThis, {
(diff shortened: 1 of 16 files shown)


jest.polyfills.js
```diff
@@ -0,0 +1,31 @@
+// jest.polyfills.js
+/**
+ * @note The block below contains polyfills for Node.js globals
+ * required for Jest to function when running JSDOM tests.
+ * These HAVE to be require's and HAVE to be in this exact
+ * order, since "undici" depends on the "TextEncoder" global API.
+ *
+ * Consider migrating to a more modern test runner if
+ * you don't want to deal with this.
```
ezr-ondrej, line 9: Huh? xD What does this mean? I guess its copied over, but do they mean more modern then Jest, or newer version of Jest?

amirfefer: I think it's because we use jsdom in jest test env, we can pick a different test environment.

ezr-ondrej: Up to you, I'll go with what you think is best solution, I was just curious ;)

amirfefer: This polyfills file  is taken from the official `msw` migration guide, so let's use it instead of replacing to a different test env :)

package.json
```diff
@@ -55,12 +57,15 @@
...
-    "msw": "1.2.3",
+    "msw": "2.0.13",
     "npm-run-all": "4.1.5",
     "prop-types": "15.8.1",
     "stylelint": "13.13.1",
     "stylelint-config-recommended-scss": "4.3.0",
     "stylelint-scss": "3.21.0",
+    "ts-patch": "^3.1.1",
+    "typescript": "^5.3.3",
+    "undici": "^5.0.0",
```
amirfefer, line 68: needs to be pin to version 5.x, there's an issue with axios and msw - https://github.com/mswjs/msw/discussions/1915

amirfefer:
tests seem to run faster :)

ezr-ondrej (approved):
Nice! :orange_heart: Great work @amirfefer, thank you ! :tada:

PR #387, opened 2024-02-28
build: bump sanitize-html

ezr-ondrej:


ezr-ondrej:
simple bump, will self-merge this one.

PR #397, opened 2024-03-26
ci(IQE-2748): Update sel testing image frontend

akhil-jha:


ezr-ondrej (approved):
sure :) I guess this will only affect stage testing, so lets get it in and lets see if it works there? :)

PR #400, opened 2024-03-28
ci: Trigger push to prod

akhil-jha:
App-interface MR 101254 is not self-serviceable and is taking time to get merged. Ideally that would trigger the pipeline as well.
I'd like to tigger the push before that.


ezr-ondrej (approved):
Sure :+1:

PR #403, opened 2024-04-01
ci: Disable parallel runs

akhil-jha:
Similar to https://github.com/RHEnVision/provisioning-backend/pull/781


ezr-ondrej (approved):
Thanks! 👍

PR #423, opened 2024-06-12
pr_check: use insights-stage as ref-env

jrusz:
This will deploy the apps using whatever is in stage which is where PRs are deployed after being merged.


adarshdubey-star (approved):

ezr-ondrej (approved):
Thanks!! :orange_heart:

PR #430, opened 2024-06-27
Bump frontend components dependencies

ezr-ondrej:
Main reason is to get rid of deprecated axios :)


ezr-ondrej:
Just a dependency bump :)

PR #432, opened 2024-06-30
build: Disable relative import lint

ezr-ondrej:
We are using babel to transform the relative pf imports, we do not need to use relative lints explicitly.

Disabling the lint is needed in that case not to force us out of the relative lints.

The transformation happens here: https://github.com/RHEnVision/provisioning-frontend/blob/408b056951f8e60cc1829d1b13b05162e3593020/babel.config.js#L44


ezr-ondrej:
@amirfefer pls let me know if you disagree, we can revert :)

PR #463, opened 2024-10-09
build: bump express dependency

ezr-ondrej:


ezr-ondrej:
Just a dep bump and only Azure test failed - known issue :)

PR #533, opened 2025-03-27
ci(HMS-5540): Use shared frontend Dockerfile

ezr-ondrej:


ezr-ondrej:
This is required to build after May 1st :) self-merging

PR #534, opened 2025-03-27
ci: add required SAST tasks to Konflux pipelines

ezr-ondrej:


ezr-ondrej:
Self merging, this is required to build after April 1st.

